PR #4, opened 2018-01-21
[Fix #1] Setup Travis CI build matrix

### LINKED ISSUE #1: Setup TravisCI with a proper build matrix
Basically we need to setup Travis in a similar way we've set it up for `cider-nrepl`, but without cljs support at this point. This includes things like code coderage, some linting, etc. 

Perhaps our resident expert @gonewest818 would be interested in helping out with that?

gonewest818:
Akin to the `cider-nrepl` setup but simpler. 

A couple of notes:

 1. I'm temporarily excluding `orchard.java.parser` from eastwood and cloverage because of `ClassNotFoundException com.sun.javadoc.ClassDoc`. Presumably the same issue that occurs in `cider-nrepl`, these third party analysis tools are running in such a way that tools.jar is not on the classpath.

2. There is a test failure only in the java 9 builds:
```
lein test :only orchard.namespace-test/project-namespaces-test
FAIL in (project-namespaces-test) (namespace_test.clj:6)
expected: (contains? (into #{} (n/project-namespaces)) (quote orchard.namespace))
  actual: (not (contains? #{} orchard.namespace))
```

3. I copied over the notifications setup from `cider-nrepl` but commented it all.  I'm not sure which are wanted, and/or what setup is required on the other end.


bbatsov:
Thanks!

> There is a test failure only in the java 9 builds:

I think that's an upstream bug, so we might have to allow this failure for now or come up with a workaround https://dev.clojure.org/jira/browse/CLASSPATH-8

I wonder why no one fixed this so far.

PR #5, opened 2018-01-23
Allow specific tests to be included/excluded based on JVM version

gonewest818:
Pull request #4 noted that there were tests failing only under Java 9.  @bbatsov recalled an upstream issue related to classloader changes to support modules in Java 9 that was probably the root cause.

For now, this set of changes implements a test selector (as in, `lein with-profile +1.8 test :java9.0`) which allows individual tests to be labeled with metadata and excluded from java9 test passes.  Same mechanism will work for future (or past) versions of Java if needs arise.

While debugging this, I noticed Travis CI doesn't support `oraclejdk7` any longer. They dropped support after Oracle dropped support. So along with the test filtering I've also reverted to `openjdk7` since that's the only Java 7 compatible JVM available in Travis CI.  I also checked and they don't yet support an `openjdk9`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,12 @@ env:
   - VERSION=1.8 TARGET=test
   - VERSION=1.9 TARGET=test
 jdk:
-  - oraclejdk7
+  - openjdk7
   - oraclejdk8
   - oraclejdk9
 matrix:
   include:
-    # Test OpenJDK against latest Clojure stable
+    # Test latest OpenJDK against latest Clojure stable
     - env: VERSION=1.9 TARGET=test
       jdk: openjdk8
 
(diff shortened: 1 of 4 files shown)


bbatsov:
Good to know! And a bit surprising given the long life-span of JDKs even after their official EOL. Clojure still support JDK 6... Anyways, we might trim the build matrix soon. :-)

gonewest818:
I see now. On the Oracle Java SE downloads page:

> Java SE 7 and Java SE 6 updates
> Updates for Java SE 7 released after April 2015, and updates for Java SE 6 released after April 2013 are only available to Oracle Customers through My Oracle Support (requires support login). 

Enterprise customers can pay for ongoing updates, but Travis CI decided not to go down that path.

PR #7, opened 2018-01-23
Port `cider.nrepl.middleware.complete`

xiongtx:
**WIP**

### DIFF
--- a/src/orchard/complete.clj
+++ b/src/orchard/complete.clj
@@ -1,0 +1,25 @@
+(ns orchard.complete
+  (:require [cljs-tooling.complete :as cljs-complete]
+            [compliment.core :as jvm-complete]
+            [compliment.utils :as jvm-complete-utils]
+            [orchard.misc :as u]
+            [orchard.util.cljs :as cljs]))
+
+(defn complete
+  [{:keys [ns symbol context extra-metadata] :as msg}]
+  (let [ns (u/as-sym ns)
+        prefix (str symbol)
+        extra-metadata (set (map keyword extra-metadata))]
+    (if-let [cljs-env (cljs/grab-cljs-env msg)]
+      (cljs-complete/completions cljs-env prefix {:context-ns ns
+                                                  :extra-metadata extra-metadata})
+      (jvm-complete/completions prefix {:ns ns
+                                        :context context
+                                        :extra-metadata extra-metadata}))))
+
+(defn complete-doc
+  [{:keys [ns symbol] :as msg}]
+  (when-not (cljs/grab-cljs-env msg)
+    (jvm-complete/documentation (str symbol) (u/as-sym ns))))
+
+(def flush-caches jvm-complete-utils/flush-caches)
--- a/src/orchard/util/cljs.clj
+++ b/src/orchard/util/cljs.clj
@@ -1,0 +1,101 @@
+(ns orchard.util.cljs)
+
+(defn try-piggieback
+  "If piggieback is loaded, returns `#'cemerick.piggieback/wrap-cljs-repl`, or
+  false otherwise."
+  []
+  (try
+    (require 'cemerick.piggieback)
+    (resolve 'cemerick.piggieback/wrap-cljs-repl)
+    (catch Exception _)))
+
+;; TODO: warn on usage of piggieback 0.1.x?
+(defn piggieback-0-2+?
+  "Returns true if piggieback 0.2.x is loaded, or false otherwise."
+  []
+  (boolean
+   (try
+     (require 'cemerick.piggieback)
+     (resolve 'cemerick.piggieback/*cljs-compiler-env*)
+     (catch Exception _))))
+
+(defn- maybe-piggieback
+  [descriptor descriptor-key]
+  (if-let [piggieback (try-piggieback)]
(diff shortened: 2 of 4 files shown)


src/orchard/util/cljs.clj
```diff
@@ -0,0 +1,101 @@
...
+    (some-> msg
+            :session
+            maybe-deref
+            (get-in path))))
+
+(defn grab-cljs-env
+  "If piggieback is active, returns the ClojureScript compiler environment for
+  the running REPL."
+  [msg]
+  (maybe-deref (grab-cljs-env* msg)))
```
xiongtx: Seems that this is making use of an nREPL message. I've yet to figure out how to untangle the CLJS stuff from the nREPL messaging--any pointers would be appreciated.

bbatsov: That's why I left this in `cider-nrepl` for now - wasn't quite clear to me what's the best way to decouple things from Piggieback, and there's also `cljs-tooling` to consider, which should likely be integrated into `orchard` itself using `cljc`. Technically the completion needs just the cljs compiler state to work, but everything piggieback related has to remain in `cider-nrepl`, as it's nREPL specific.

src/orchard/complete.clj
```diff
@@ -0,0 +1,25 @@
+(ns orchard.complete
+  (:require [cljs-tooling.complete :as cljs-complete]
+            [compliment.core :as jvm-complete]
+            [compliment.utils :as jvm-complete-utils]
+            [orchard.misc :as u]
+            [orchard.util.cljs :as cljs]))
+
+(defn complete
```
bbatsov: The API of those functions has to be changed to be non-nREPLy. For normally callers something like a `msg` doesn't make much sense here and probably this can just have normal positional arguments.

src/orchard/complete.clj
```diff
@@ -0,0 +1,25 @@
...
+  [{:keys [ns symbol context extra-metadata] :as msg}]
+  (let [ns (u/as-sym ns)
+        prefix (str symbol)
+        extra-metadata (set (map keyword extra-metadata))]
+    (if-let [cljs-env (cljs/grab-cljs-env msg)]
```
bbatsov: I guess to make this portable outside nREPL we should use `cljc`to check whether we're dealing with a Java or JavaScript completion.

src/orchard/complete.clj
```diff
@@ -0,0 +1,25 @@
+(ns orchard.complete
+  (:require [cljs-tooling.complete :as cljs-complete]
```
bbatsov: This entire library has to be integrated into `orchard`, as it was developed for a similar purpose and it got just to two features - completion and var-info.

bbatsov:
Generally for completions there's very little logic in `cider-nrepl` - it's mostly deferring to two other libs. To make this useful here we ideally need a universal completion function that works in both Clojure and ClojureScript.

src/orchard/complete.clj
```diff
@@ -0,0 +1,25 @@
...
+      (cljs-complete/completions cljs-env prefix {:context-ns ns
+                                                  :extra-metadata extra-metadata})
+      (jvm-complete/completions prefix {:ns ns
+                                        :context context
+                                        :extra-metadata extra-metadata}))))
+
+(defn complete-doc
+  [{:keys [ns symbol] :as msg}]
+  (when-not (cljs/grab-cljs-env msg)
+    (jvm-complete/documentation (str symbol) (u/as-sym ns))))
```
bbatsov: Another missing feature in the cljs version, which makes this function useless outside `cider-nrepl` in its current form.

src/orchard/complete.clj
```diff
@@ -0,0 +1,25 @@
...
+
+(def flush-caches jvm-complete-utils/flush-caches)
```
bbatsov: That's just an alias. Not see what's the point in having it here.

src/orchard/complete.clj
xiongtx: Sure, but I wasn't sure how to remove the `msg` due to the `(cljs/grab-cljs-env msg)`.

src/orchard/complete.clj
xiongtx: To avoid the client having to invoke `compliment.utils/flush-caches` directly. They'd have to require the `compliement.utils/flush-caches` ns otherwise.

src/orchard/complete.clj
bbatsov: On the other hand - I'm assuming that with piggieback we'd actually be in `clj`, so probably this has to be some optional param we could pass (`cljs-env`).

bbatsov:
After thinking a bit more what to do about this I think probably the best approach would be to:

* move cljs-tooling to orchard
* provide some unified completion api that cider-nrepl can use (although this leave open the question how to handle things that existing only in compliment right now like completion caches) 
* leave the piggieback stuff to cider-nrepl, as they are nrepl-specific

Ideally we should get to a point where cider-nrepl doesn't know (much) about the underlying completion provider.

bbatsov:
And I guess something very similar should be done for var-info, as cljs-tooling just tried to emulate the original cider-nrepl api.

PR #10, opened 2018-01-25
[Fix #9] auto publish snapshots to clojars

### LINKED ISSUE #9: Auto-publish snapshots to Clojars
Splitting out this request from the discussion in #2:

> I was wondering if you know how to setup some auto-publishing of snapshots to Travis. I find it very annoying that I manually have to push each update. Probably the only real problem is how to stuff some credentials securely in the build step. Would be nice for both orchard and cider-nrepl if possible and would get rid of the common problem of forgetting to push some new snapshot release.

gonewest818:
This publish should achieve auto-publish to clojars for snapshots, and controlled publishes for releases too.

The changes are as follows.

(1) In `project.clj`, add a `:deploy-repositories` configuration for clojars.  Note that the clojars credentials do need to be set in the environment, but GPG signing is disabled because as of today there is no mechanism for unattended GPG signing in Leiningen (see https://github.com/technomancy/leiningen/pull/2279). Secure credentials set manually in the Travis CI "Settings" page will be elided from the build logs like this:
```
Setting environment variables from repository settings
$ export CLOJARS_USERNAME=[secure]
$ export CLOJARS_PASSWORD=[secure]
```

(2) In `project.clj` add a `:release-tasks` that performs manipulation of the project version string, adds a tag, and commits changes to git. After a `lein release` you'll see two new commits. First commit has the description "Release 0.1.0" and is tagged "v0.1.0". Second commit is "Begin 0.1.1-SNAPSHOT". Pushing these commits will cause Travis CI jobs to run, and it's in the Travis CI job where the actual deploys will be done.

```
$ lein release 
[9-auto-publish-snapshots 4a90d89] Release 0.1.0
 1 file changed, 1 insertion(+), 1 deletion(-)
[9-auto-publish-snapshots 2ee967c] Begin 0.1.1-SNAPSHOT
 1 file changed, 1 insertion(+), 1 deletion(-)
```

(3) In the `Makefile` add targets `make release` and `make deploy`.  These are equivalent to the Leiningen commands, used in the Travis CI config.

(4) `.travis.yml` is ported from "build matrix" to "build stages" which is a beta feature. In this new style of config the `stages:` map defines three sequential stages that will run only if the prior stages finish successfully. (*)It's worth noting the `allow_failures:` field still applies here and any jobs allowed to fail can fail without blocking the deploy.

As in the prior `.travis.yml`, there is matrix of jobs implied by the list of `jdk:` and `env:`. This matrix is enumerated as before, and will be run as part of the "test" stage. For clarity I've moved the eastwood and cljfmt jobs into a separate "check" stage. The "deploy" stage is last.

The net result of all this is, any commit you push to the master branch will result in a new artifact on clojars (provided the CI jobs succeed).  If you happen to be working on a snapshot version then a new snapshot is published, and if you happen to have run `make release` then you'll get a new release version AND the first iteration of the next snapshot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,12 @@ jdk:
   - openjdk7
   - oraclejdk8
   - oraclejdk9
-matrix:
+stages:
+  - name: test
+  - name: check
+  - name: deploy
+    if: branch = master
+jobs:
   include:
     # Test latest OpenJDK against latest Clojure stable
     - env: VERSION=1.9 TARGET=test
@@ -24,18 +29,28 @@ matrix:
     - env: VERSION=master TARGET=test
       jdk: oraclejdk8
 
-    # Eastwood linter
-    - env: VERSION=1.9 TARGET=eastwood
-      jdk: oraclejdk8
-
-    # Check cljfmt
-    - env: VERSION=1.9 TARGET=cljfmt
-      jdk: oraclejdk8
-
     # Coverage analysis
     - env: VERSION=1.9 TARGET=cloverage
       jdk: oraclejdk8
-      after_success: bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
+      after_success:
+      - >
+        test $TARGET = "cloverage" &&
+        bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
+
+    # Eastwood linter
+    - stage: check
+      env: VERSION=1.9 TARGET=eastwood
+      jdk: oraclejdk8
+
+    # Check cljfmt
+    - stage: check
+      env: VERSION=1.9 TARGET=cljfmt
+      jdk: oraclejdk8
+
+    # Deployment
+    - stage: deploy
+      env: TARGET=deploy-xxx-dont-actually
+      jdk: oraclejdk8
 
   fast_finish: true      # don't wait for allowed failures before build finish
   allow_failures:
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean test eastwood cljfmt cloverage
+.PHONY: clean test eastwood cljfmt cloverage release deploy
 
 VERSION ?= 1.9
 
@@ -28,6 +28,22 @@ cljfmt:
 cloverage:
 	lein with-profile +$(VERSION),+cloverage cloverage --codecov \
 	     -e "orchard.java.parser"
+
+# When releasing, the BUMP variable controls which field in the
+# version string will be incremented in the *next* snapshot
+# version. Typically this is either "major", "minor", or "patch".
+
+BUMP ?= patch
+
+release:
+	lein with-profile +$(VERSION) release $(BUMP)
+
+# Deploying requires the caller to set environment variables as
+# specified in project.clj, to provide a login and password to the
+# artifact repository
+
+deploy:
+	lein with-profile +$(VERSION) deploy clojars
 
 clean:
 	lein clean
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 [![Build Status](https://travis-ci.org/clojure-emacs/orchard.png?branch=master)](https://travis-ci.org/clojure-emacs/orchard)
 [![Dependencies Status](https://versions.deps.co/clojure-emacs/orchard/status.svg)](https://versions.deps.co/clojure-emacs/orchard)
+[![Coverage](https://codecov.io/gh/clojure-emacs/orchard/branch/master/graph/badge.svg)](https://codecov.io/gh/clojure-emacs/orchard/)
+
 
 # orchard (alpha)
 
--- a/project.clj
+++ b/project.clj
@@ -14,6 +14,20 @@
 
   :test-selectors {:java9.0 (complement :java9-excluded)}
 
+  :aliases {"bump-version" ["change" "version" "leiningen.release/bump-version"]}
+
+  :release-tasks [["vcs" "assert-committed"]
+                  ["bump-version" "release"]
+                  ["vcs" "commit" "Release %s"]
+                  ["vcs" "tag" "v" "--no-sign"]
+                  ["bump-version"]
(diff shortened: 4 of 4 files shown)


.travis.yml
```diff
@@ -14,7 +14,12 @@ jdk:
   - openjdk7
   - oraclejdk8
   - oraclejdk9
-matrix:
+stages:
+  - name: test
+  - name: check
+  - name: deploy
+    if: branch = master
```
gonewest818: We need to test if this is enough to suppress pull requests from attempting a deploy. I'm just not sure.

I'm not worried about bad jars being deployed, because Travis CI makes sure pull requests don't inherit your environment variable settings, therefore won't be able to publish jars to your clojars account... but what I don't want is for every pull request to be labeled "Checks have failed" because of the deploy step.

bbatsov:
Looks great! The only thing I need to do now is setup some dedicated account for this, as many people are admin of the `clojure-emacs` org and I obviously don't want to share my personal Clojars credentials with them. :-)

gonewest818:
fwiw, if creating another account is a hassle you'd rather not deal with, then the other way to handle this is to encrypt those strings with the `travis` command line tool.  That results in something you can safely put in `.travis.yml`, provided you trust in the crypto.

See https://docs.travis-ci.com/user/encryption-keys/

That'll end up looking something like
```
env:
  global:
    - secure: mcUCykGm4bUZ3CaW6... etc
```

Where the crypto hides the name of the environment variable as well as its value, and is unpacked into the build but elided from the logs, etc.

bbatsov:
I like this idea. You truly are a Travis master! :-) I've used it casually for many years, never really learned much about it.

gonewest818:
Not at all. Really it’s experimentation, which in Travis means commit and wait, commit and wait, ... You should see what my commit log looked like before I reordered and squashed commits for the PR.

Anyway, now is you chance — you’ll have to run the `Travis` cli tool to generate the encrypted secrets because only you know your password. Let me know how that goes.

bbatsov:
Done. Let's hope I didn't mess something up. :-) Next step - do the same for `cider-nrepl` (if it works).

gonewest818:
Nice. When you put in the encrypted environment variables, you're going to need to edit `.travis.yml`  where it says `env: TARGET=deploy-xxx-dont-actually` because I left that toggled off.

bbatsov:
Done.

gonewest818:
...my mistake, sorry, I forgot I had done that out of paranoia in my own testing.

gonewest818:
I'm looking at the [failed Travis CI job](https://travis-ci.org/clojure-emacs/orchard/jobs/333829315). What I can see in the logs looks ok.

`CLOJARS_USERNAME` and `CLOJARS_PASSWORD` are in the environment:

screen shot 2018-01-26 at 9 37 10 am

and `make` is invoking the deploy as expected. The failure is the part about "unauthorized":

screen shot 2018-01-26 at 9 57 23 am

There are some caveats in the Travis CI docs about quoting the environment variables, for example if the password contains characters like `#` that might be interpreted by the shell... ?

bbatsov:
> There are some caveats in the Travis CI docs about quoting the environment variables, for example if the password contains characters like # that might be interpreted by the shell

It's just letters and digits. I get exactly the same mistake locally with `make deploy`, having exported those variables locally. Seems the mistake is elsewhere, right?

gonewest818:
Right. It’s as if the environment variables aren’t passed by make into lein, which is certainly possible. I’ll go take a look.

gonewest818:
Made some progress on this...

I copied the .travis.yml setup into [lein-sysutils](https://github.com/gonewest818/lein-sysutils), a project I created earlier today. The first Travis CI deploy stalled for 10 minutes before aborting. I hadn't set a `CLOJARS_USERNAME` and `CLOJARS_PASSWORD` at all, and in that case Leiningen prompts for input and waits. Travis aborts the job after a timeout of 10 minutes.

For the next few builds, I used the web "settings" page in Travis CI and entered the credentials there:

screen shot 2018-01-27 at 8 35 03 pm

At first, these deploys failed too... because my particular password does not work without escaping. But, the deploy jobs failed fast with the "unauthorized" message, no more 10 minute timeouts, giving me confidence that the environment variables were getting passed through `make` into the `lein deploy`, but the values `lein` received weren't right.

So then I finally single quoted my password in the web UI, and I got a successful deploy. Here's a log: https://travis-ci.org/gonewest818/lein-sysutils/jobs/334275869

and here's Clojars: https://clojars.org/lein-sysutils

Next step, I'll try encrypting the environment variables with the `travis` CLI.

gonewest818:
....and I can successfully deploy with encrypted environment variables in `.travis.yml`, having deleted the manually entered environment variables from the Travis CI "Settings" web page.

Ultimately what I had to do was single quote the entire string (`travis encrypt 'VARIABLE=value'`) and escape with `\` any characters in the `value` that could be interpreted by bash.  And that worked.

The `.travis.yml`: https://github.com/gonewest818/lein-sysutils/blob/master/.travis.yml#L14-L16

Deploy log: https://travis-ci.org/gonewest818/lein-sysutils/jobs/334285613

and Clojars repo, now with a second snapshot from this deploy: https://clojars.org/repo/lein-sysutils/lein-sysutils/0.1.0-SNAPSHOT/

Granted, this is all in a different github project but the Makefile and Travis CI configurations are very very similar by design. I'm fairly confident in the setup now, wondering if it's something in the encryption or escaping.

bbatsov:
@gonewest818 But I'm really puzzled why this doesn't work for me locally. Did you have trouble running `make deploy` at all?

gonewest818:
No problems … `make deploy` works for me locally, provided I set those environment variables correctly in the shell.

bbatsov:
I found the problem and it's absurd - I used the username `bbatsov` most of the time, but on Clojars I'm `bozhidar`. Guess which username I had encrypted? 😆

gonewest818:
Glad to see it working!

PR #11, opened 2018-01-25
fix two broken tests

gonewest818:
The `inspect-val-test` was throwing:
```
Exception in thread "main" java.lang.RuntimeException: Unable to resolve symbol: inspect-value in this context, compiling:(orchard/inspect_test.clj:208:1)
```

The `map-structure-test` error was 
```
FAIL in (map-structure-test) (java_test.clj:41)
Parsed map structure = reflected map structure
expected: (keys= c1 c2)
  actual: false
```

and I freely admit I don't understand the internals of this, but I inspected the contents of c1 and c2 and discovered the difference was the `:path` key.


bbatsov:
> Exception in thread "main" java.lang.RuntimeException: Unable to resolve symbol: inspect-value in this context, compiling:(orchard/inspect_test.clj:208:1)

I messed this up when I moved it yesterday.

> and I freely admit I don't understand the internals of this, but I inspected the contents of c1 and c2 and discovered the difference was the :path key.

The new key was recently added, but I guess we didn't notice it broken a test. 

Thanks a lot! 👍

PR #13, opened 2018-01-28
[#9] final cleanup to Travis CI config

gonewest818:
Two small changes that keep this repo in line with `cider-nrepl`

 - the `after_success` script in the `cloverage` task needlessly checked the value of `$(TARGET)`
 - `make deploy` would attempt (and fail) to deploy from a pull request job.  This does a more proper job of catching that.


bbatsov:
👍

PR #15, opened 2018-02-12
use lein-sysutils to clean up java version detection

gonewest818:
Like we did in clojure-emacs/cider-nrepl#501.


bbatsov:
Hmm, seems the codecov integration is finally working!

PR #16, opened 2018-02-14
[clojure-emacs/cider-nrepl#489] Migrate tests from cider-nrepl

gonewest818:
This is the companion to clojure-emacs/cider-nrepl#507 in which test namespaces were restructured to eliminate orchard dependencies. In a few cases, we needed to migrate tests into orchard.


bbatsov:
Nice! 👍

gonewest818:
Thank you.

PR #22, opened 2018-03-07
Return apropos type of multifn as function

xiongtx:
`MultiFn`s are not `Fn`, but for the purpose of apropos they should be treated as such.

### DIFF
--- a/src/orchard/apropos.clj
+++ b/src/orchard/apropos.clj
@@ -3,7 +3,8 @@
   {:author "Jeff Valk"}
   (:require [orchard.meta :refer [var-name var-doc]]
             [orchard.meta :as m]
-            [orchard.namespace :as ns]))
+            [orchard.namespace :as ns])
+  (:import [clojure.lang MultiFn]))
 
 ;;; ## Overview
 ;;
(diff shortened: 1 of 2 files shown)


arrdem:
Hummmm do we actually want that `or`? Wouldn't we actually want to test for `ifn?`?

xiongtx:
>  Wouldn't we actually want to test for `ifn?`?

No; if the value of a var is a map, this would classify it as a function.

```clojure
(def foo {:a 1 :b 2})
(ifn? foo)    ;; => true
```

arrdem:
Hummmmm from Clojure's perspective `AFn` is really the class you care about from the perspective of `instanceof` if you want to differentiate "actual" functions from "data with operational semantics" as it were, but it's not great. I'd go so far as to argue that it is correct to say `{}` is a two-arity function. I've seen enough code which used sets and maps respectively as predicates and pure functions.

xiongtx:
> Hummmmm from Clojure's perspective `AFn` is really the class you care about from the perspective of instanceof if you want to differentiate "actual" functions from "data with operational semantics"

```clojure
(instance? clojure.lang.AFn {:a 1 :b 2})    ;; => true
```

> I'd go so far as to argue that it is correct to say `{}` is a two-arity function.

I don't think most users would like to `(def foo {:a 1 :b 2})` and have `apropos` report it as a function instead of variable.

bbatsov:
I like the change, I just wonder if it should not get an unit test.

xiongtx:
Added some tests

PR #23, opened 2018-03-16
Make sure project namespaces are found on Windows* OSes

benedekfazekas:
When filtering the classpath with the project root lowercase both
sides of the comparison if the OS is Windows.

Enables running all tests on Windows, without this change no tests are
found because of a lower/upper case mismatch when filtering
the project neses on Windows.

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/orchard/misc.clj
+++ b/src/orchard/misc.clj
@@ -1,6 +1,12 @@
 (ns orchard.misc
   (:require [clojure.string :as str]))
 
+(def ^:const windows-prefix
+  "Windows")
+
+(defn os-windows? []
+  (.startsWith (System/getProperty "os.name") windows-prefix))
+
 (defn boot-fake-classpath
   "Retrieve Boot's fake classpath.
   When using Boot, fake.class.path contains the original directories with source
--- a/src/orchard/namespace.clj
+++ b/src/orchard/namespace.clj
@@ -4,7 +4,9 @@
             [clojure.tools.namespace.file :as ns-file]
             [clojure.tools.namespace.find :as ns-find]
             [orchard.classloader :refer [class-loader]]
-            [orchard.classpath :as cp])
+            [orchard.classpath :as cp]
+            [clojure.string :as str]
+            [orchard.misc :as misc])
   (:import java.util.jar.JarFile))
 
 ;;; Namespace Loading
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. This change
 ## [Unreleased]
 ### Changed
 - Add a new arity to `make-widget-async` to provide a different widget shape.
+- Make sure project namespaces are found under Windows* type operating systems.
```
bbatsov: What's this file?

src/orchard/misc.clj
```diff
@@ -1,6 +1,12 @@
 (ns orchard.misc
   (:require [clojure.string :as str]))
 
+(def ^:const windows-prefix
+  "Windows")
+
+(def ^:const os-windows?
```
bbatsov: I'm not a big fan of this approach as well - why not make this a function? I doubt the performance impact would be meaningful.

src/orchard/namespace.clj
```diff
@@ -32,9 +34,15 @@
 (defn project-namespaces
   "Find all namespaces defined in source paths within the current project."
   []
-  (->> (filter (memfn isDirectory) (cp/classpath (class-loader)))
-       (filter #(.startsWith (str %) project-root))
-       (mapcat ns-find/find-namespaces-in-dir)))
+  (let [project-pred (or
+                      (and misc/os-windows?
+                           #(.startsWith
```
bbatsov: I'd add some comment here explaining the need for this.

src/orchard/misc.clj
benedekfazekas: can turn into an fn, noprobs. but it won't ever change obviously, or? ;)

benedekfazekas:
still working on this!!

benedekfazekas:
added test. ready on my side

src/orchard/misc.clj
bbatsov: My main issue with those constants is that they are hard to test. But it's not a big deal for me.

benedekfazekas: indeed. changed it to fn and added a test

benedekfazekas:
cheers!

PR #24, opened 2018-03-20
Improve inspector

alexander-yakushev:
Two quite major things coming to the inspector, long overdue.

### Separate key/value rendering in maps

We had this long time ago, somehow it was lost, now it's time to bring it back. Having to click twice when navigating maps is really annoying. Besides, this behavior was already implemented for metadata - go figure.

As a nice bonus, the path tracking code is now simpler and more robust.

### Separate page numbers for each level of nesting

This was a blatant long-standing bug. By having a single value to represent the current page we lose information when we start navigating up and down the data tree. Page index of the parent collection becomes the page of the child collection. This is very confusing. This PR gives each nesting level its own page number by maintaining a stack of them.

----

The changes should be fully compatible with CIDER and cider-nrepl – nothing to change there besides, maybe, a few integration tests.


bbatsov:
Looks great to me! Just two small things:

* It'd be nice to add changelog entries for those fixes
* While you're at it - can you you also check what caused https://github.com/clojure-emacs/cider/issues/2152?

Some tests on the cider-nrepl side would be appreciated, but are not mandatory.

alexander-yakushev:
I planned to edit the cider-nrepl part (including tests) once this is merged and an artifact is created. What is the preferred way of co-developing orchard and cider-nrepl?

I will add to the changelog, sure.

alexander-yakushev:
Regarding #2152, I suffer from this issue too. I was unable to fix it on the Orchard side, but managed to solve it on the cider-nrepl. I will submit that fix together with cider-nrepl tests for this set of changes.

bbatsov:
> I planned to edit the cider-nrepl part (including tests) once this is merged and an artifact is created. What is the preferred way of co-developing orchard and cider-nrepl?

Developing things sequentially is fine.

bbatsov:
Merging this PR will immediately deploy a new snapshot to Clojars.

PR #25, opened 2018-05-06
Add environment query functions

SevereOverfl0w:
This was added to allow the sharing of query logic between various
different facets of cider.  Apropos, tests, and the namespace explorer
all expose different methods for operating on sets of vars from the
environment.

In particular, I want to do an apropos of only project namespaces, and
this isn't possible with current filtering rules.

I'm suspect that my naming is not very good, so feedback on that is good.  It's
also notable that this changes apropos to use clojure types natively, and
expects cider-nrepl to do the correct coercion from bencode.

### DIFF
--- a/src/orchard/apropos.clj
+++ b/src/orchard/apropos.clj
@@ -1,9 +1,8 @@
 (ns orchard.apropos
   "Search symbols and docs matching a regular expression"
   {:author "Jeff Valk"}
-  (:require [orchard.meta :refer [var-name var-doc]]
-            [orchard.meta :as m]
-            [orchard.namespace :as ns])
+  (:require [orchard.meta :refer [var-name var-doc] :as m]
+            [orchard.query :as query])
   (:import [clojure.lang MultiFn]))
 
 ;;; ## Overview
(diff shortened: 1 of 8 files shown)


src/orchard/query_environment.clj
```diff
@@ -0,0 +1,67 @@
+(ns orchard.query-environment
```
bbatsov: Perhaps `query-repl`, `repl-state` or `inspect-repl`? I'm afraid for many people "environment" might mean a shell for instance. 

The namespace should also have some docstring.

src/orchard/query_environment.clj
```diff
@@ -0,0 +1,67 @@
...
+  (:require [orchard.meta :as m]
+            [orchard.namespace :as ns]))
+
+(defn list-ns
```
bbatsov: All defns should be documented, otherwise the users will really struggle to figure out what to do with them.

src/orchard/apropos.clj
```diff
@@ -15,43 +14,41 @@
...
+          :else (compare (str x) (str y))))
+      vars)))
 
 (defn find-symbols
   "Find symbols or (optionally) docstrings matching `query` in `search-ns` if
   specified or all namespaces. The search may optionally include private vars,
   and may be case senstive. Types returned correspond to Apropos types.
   Docstring search returns the full doc; symbol search returns an abbreviated
   version."
-  [{:keys [ns query search-ns docs? privates? case-sensitive? filter-regexps]}]
```
bbatsov: You've changed the signature of this function, but you didn't change its docstring.

src/orchard/meta.clj
```diff
@@ -146,6 +146,12 @@
   (concat (keys (var-get #'clojure.repl/special-doc-map))
           '[& catch finally]))
 
+(defn meta+
+  "Return special form or var's meta"
```
bbatsov: You forgot the final `.` here.

src/orchard/meta.clj
```diff
@@ -146,6 +146,12 @@
...
+  [v]
+  (or (special-sym-meta v)
```
bbatsov: Hmm, I see we don't use this in `var-meta`. Perhaps its docstring is out of date and this used to be the same function.

src/orchard/meta.clj
```diff
@@ -146,6 +146,12 @@
...
+      (meta v)))
```
bbatsov: `meta` or `var-meta`?

src/orchard/namespace.clj
```diff
@@ -127,3 +127,8 @@
               (first paths)
               (recur (rest paths)))))))
     (catch Throwable _ nil)))
+
+(defn has-tests?
+  "Return a truthy value if the namespace has any `:test` metadata."
```
bbatsov: Perhaps "namespace vars have..." would make it more clear what this actually does. :-)

src/orchard/query_environment.clj
SevereOverfl0w: I'm not keen on the name either, I'll look for something existing. REPL has too many connotations with the command line prompt, and not enough with clojure's state.

SevereOverfl0w:
I imported the official eastwood config into the project, which should remove the linting failure.

bbatsov:
👍

PR #29, opened 2018-09-20
Remove reflection warning

slipset:
The Clojure compiler is unable to devine that `partitioned` is a `clojure.lang.LazySeq`, and emits a reflection warning when Cider is run in a project where `*warn-on-reflection*` is set to `true` globally.
This type-hint removes that warning

- [ x] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [x ] All tests are passing
- [ x] The new code is not generating reflection warnings 

Thanks!

### DIFF
--- a/src/orchard/eldoc.clj
+++ b/src/orchard/eldoc.clj
@@ -80,7 +80,7 @@
                  (or (:in query) "$")
                  ;; query as vector
                  (let [partitioned (partition-by keyword? query)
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm surprised that's our only warning - I rarely check for those, so I assumed there would be more. :-)

vspinu:
There are tons of reflexion warnings actually. Many of them are in the server buffer.

PR #35, opened 2018-11-06
Various inspector improvements

alexander-yakushev:
A few useful goodies. Nothing too ground-breaking, so this might still have a chance to land into the upcoming cider-nrepl release.

- Render everything that implements j.u.List and j.u.Map as collections –– useful when interacting with Java libraries.
- Render inline arrays as collections – same.
- Truncate inline long strings and other objects – this helps when the inspected object has long strings as fields/contained values, or also unknown objects with toString implementations that produce long strings. The truncation limit is quite arbitrary, but I don't think this would be too useful to customize. Also, jumping into the string still renders it fully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+* [#35](https://github.com/clojure-emacs/orchard/pull/35): [Inspector] Render Java's lists, maps, and arrays as collections.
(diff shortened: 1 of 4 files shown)


alexander-yakushev:
Also bumped the version to snapshot because tests are failing :).

bbatsov:
Great work!

Issue #20, opened 2018-02-23
Figure out what to do with our dynapath 1.0 compatibility problems


bbatsov:
Long story short:

* We don't support the latest version of dynapath (1.0)
* The version we're using (0.2.5) is going to stop working with Java 10, as it relied on implementation details that are going to be changed

We need to come up with some solution for this soon, otherwise it's going to cause problems for many people. 

More details here https://github.com/clojure-emacs/cider-nrepl/issues/482

tsulej:
Maybe it's related, I've just checked Java 10 and got such warning. nRepl and Cider seem to work.

nREPL server started on port 51299 on host 127.0.0.1 - nrepl://127.0.0.1:51299
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by mranderson048.orchard.v0v1v0_20180321v113210_28.dynapath.v0v2v5.dynapath.defaults$eval7485$fn__7486 to method java.net.URLClassLoader.addURL(java.net.URL)
WARNING: Please consider reporting this to the maintainers of mranderson048.orchard.v0v1v0_20180321v113210_28.dynapath.v0v2v5.dynapath.defaults$eval7485$fn__7486
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
REPL-y 0.3.7, nREPL 0.2.12
Clojure 1.9.0
Java HotSpot(TM) 64-Bit Server VM 10+46

SevereOverfl0w:
Mirroring the boot approach seems difficult unless we control the start of the JVM very closely. I believe it needs to run prior to clojure.

bbatsov:
Unfortunately I'm out of other ideas. I'm guessing something must be possible, but I simply don't have the time to do the research for it. The fact that dynapath decided to drop this functionality instead of fixing it leads me to believe probably there's no easy solution, though.

SevereOverfl0w:
Inside of boot, I think we're okay, as we can use their compatible classloader.

What features does dynapath provide to us?

bbatsov:
We use it only to add the JDK resources dynamically to the classpath, so afterwards things like source lookup and doc lookup will work for them. See https://github.com/clojure-emacs/orchard/blob/3934ea63117b3cd3f1b459e66ada05ac173a1fc1/src/orchard/java.clj#L32

Unfortunately we're using exactly the damn method that the deprecation warnings are about. :-)

[...]

PR #37, opened 2018-12-30
Port cljs info and consolidate info api

arichiardi:
In order to start consolidating Clojure and ClojureScript tooling under our apple orchard, this patch starts porting the info op from cljs-tooling. Consequently, it paves the way to port some of the orchard Clojure-only code to ClojureScript.

- [x] All tests are passing
- [X] The new code is not generating reflection warnings

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -4,7 +4,8 @@
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :scm {:name "git" :url "https://github.com/clojure-emacs/orchard"}
-  :dependencies [[org.tcrawley/dynapath "1.0.0"]]
+  :dependencies [[org.tcrawley/dynapath "1.0.0"]
+                 [org.clojure/clojurescript "1.10.520"]]
   :exclusions [org.clojure/clojure] ; see versions matrix below
 
   :aliases {"bump-version" ["change" "version" "leiningen.release/bump-version"]}
@@ -37,6 +38,16 @@
                                      [org.clojure/clojure "1.11.0-master-SNAPSHOT" :classifier "sources"]]}
 
              :sysutils {:plugins [[lein-sysutils "0.2.0"]]}
+
+             :test {:resource-paths ["test-resources"]}
+
+             ;; DEV tools
+             :dev {:dependencies [[pjstadig/humane-test-output "0.9.0"]]
+                   :resource-paths ["test-resources"]
+                   :plugins [[com.jakemccrary/lein-test-refresh "0.23.0"]]
+                   :injections [(require 'pjstadig.humane-test-output)
+                                (pjstadig.humane-test-output/activate!)]
+                   :test-refresh {:changes-only true}}
 
              ;; CI tools
              :cloverage {:plugins [[lein-cloverage "1.0.11-SNAPSHOT"]]}
--- a/src/orchard/classpath.clj
+++ b/src/orchard/classpath.clj
@@ -7,6 +7,7 @@
   (:import
    (java.io File)
    (java.net URI URL)
+   (java.nio.file Paths)
    (java.util.jar JarFile JarEntry)))
 
 ;;; Classloaders
@@ -84,3 +85,23 @@
       (->> (file-seq f)
            (filter #(not (.isDirectory ^File %)))
            (map #(.getPath (.relativize (.toURI url) (.toURI ^File %))))))))
+
+;;; Boot's hack - previously part of cider-nrepl
+
+(defn classpath-file-relative-path
+  "Boot stores files in a temporary directory & ClojureScript stores
+  the :file metadata location absolutely instead of relatively to the
+  classpath. This means when doing jump to source in Boot &
+  ClojureScript, you end up at the temp file.  This code attempts to
+  find the classpath-relative location of the file, so that it can be
+  opened correctly."
+  [s]
+  (let [path (Paths/get s (into-array String []))
+        path-count (.getNameCount path)]
+    (or (first
+         (sequence
+          (comp (map #(.subpath path % path-count))
+                (map str)
+                (filter io/resource))
+          (range path-count)))
+        s)))
--- a/src/orchard/cljs/analysis.cljc
+++ b/src/orchard/cljs/analysis.cljc
@@ -5,11 +5,9 @@
   (:require [orchard.misc :as u])
   (:refer-clojure :exclude [find-ns find-var all-ns ns-aliases]))
 
-(def NSES :cljs.analyzer/namespaces)
-
 (defn all-ns
   [env]
-  (->> (NSES env)
+  (->> (:cljs.analyzer/namespaces env)
        ;; recent CLJS versions include data about macro namespaces in the
        ;; compiler env, but we should not include them in completions or pass
        ;; them to format-ns unless they're actually required (which is handled
@@ -121,8 +119,8 @@
                  (filter macro?)
                  (into {})))
      :cljs (->> (merge
-                 (get-in env [NSES ns :macros])
-                 (get-in env [NSES ns :defs]))
+                 (get-in env [:cljs.analyzer/namespaces ns :macros])
+                 (get-in env [:cljs.analyzer/namespaces ns :defs]))
                 (remove (fn [[k v]] (:private v)))
                 (into {}))))
 
@@ -148,8 +146,8 @@
   [env ns]
   {:pre [(symbol? ns)]}
   (merge
-   (get-in env [NSES ns :macros])
-   (get-in env [NSES ns :defs])))
+   (get-in env [:cljs.analyzer/namespaces ns :macros])
+   (get-in env [:cljs.analyzer/namespaces ns :defs])))
 
 (defn sanitize-ns
   "Add :ns from :name if missing."
--- a/src/orchard/cljs/analysis.cljc
+++ b/src/orchard/cljs/analysis.cljc
@@ -1,0 +1,206 @@
+(ns ^{:doc "ClojureScript analysis functions."
+      :author "Gary Trakhman"
+      :added "0.6.0"}
+ orchard.cljs.analysis
+  (:require [orchard.misc :as u]
+            #?(:clj [cljs.repl]))
+  (:refer-clojure :exclude [find-ns find-var all-ns ns-aliases]))
+
+(def NSES :cljs.analyzer/namespaces)
+
+(defn all-ns
+  [env]
+  (->> (NSES env)
+       ;; recent CLJS versions include data about macro namespaces in the
+       ;; compiler env, but we should not include them in completions or pass
+       ;; them to format-ns unless they're actually required (which is handled
+       ;; by macro-ns-candidates below)
+       (into {} (filter (fn [[_ ns]]
+                          (not (and (contains? ns :macros)
+                                    (= 1 (count ns)))))))))
+
+(defn find-ns
+  [env ns]
+  (get (all-ns env) ns))
+
+;; Code adapted from clojure-complete (http://github.com/ninjudd/clojure-complete)
+
+(defn imports
+  "Returns a map of [import-name] to [ns-qualified-import-name] for all imports
+  in the given namespace."
+  [env ns]
+  (:imports (find-ns env ns)))
+
+(defn ns-aliases
+  "Returns a map of [ns-name-or-alias] to [ns-name] for the given namespace."
+  [env ns]
+  (let [imports (imports env ns)]
+    (->> (find-ns env ns)
+         :requires
+         (filter #(not (contains? imports (key %))))
+         (into {}))))
+
+(defn macro-ns-aliases
+  "Returns a map of [macro-ns-name-or-alias] to [macro-ns-name] for the given namespace."
+  [env ns]
+  (:require-macros (find-ns env ns)))
+
+(defn- expand-refer-map
+  [m]
+  (into {} (for [[k v] m] [k (symbol (str v "/" k))])))
+
+(defn referred-vars
+  "Returns a map of [var-name] to [ns-qualified-var-name] for all referred vars
+  in the given namespace."
+  [env ns]
+  (->> (find-ns env ns)
+       :uses
+       expand-refer-map))
+
+(defn referred-macros
+  "Returns a map of [macro-name] to [ns-qualified-macro-name] for all referred
+  macros in the given namespace."
+  [env ns]
+  (->> (find-ns env ns)
+       :use-macros
+       expand-refer-map))
+
+(defn ns-alias
+  "If sym is an alias to, or the name of, a namespace referred to in ns, returns
+  the name of the namespace; else returns nil."
+  [env sym ns]
+  (get (ns-aliases env ns) sym))
+
+(defn macro-ns-alias
+  "If sym is an alias to, or the name of, a macro namespace referred to in ns,
+  returns the name of the macro namespace; else returns nil."
+  [env sym ns]
+  (get (macro-ns-aliases env ns) sym))
+
+(defn- public?
+  [[_ var]]
+  (not (:private var)))
+
+(defn- named?
+  [[_ var]]
+  (not (:anonymous var)))
+
+(defn- foreign-protocol?
+  [[_ var]]
+  (and (:impls var)
+       (not (:protocol-symbol var))))
+
+(defn- macro?
+  [[_ var]]
+  (:macro (meta var)))
+
+(defn ns-vars
+  "Returns a list of the vars declared in the ns."
+  [env ns]
+  (->> (find-ns env ns)
+       :defs
+       (filter (every-pred named? (complement foreign-protocol?)))
+       (into {})))
+
+(defn public-vars
+  "Returns a list of the public vars declared in the ns."
+  [env ns]
+  (->> (find-ns env ns)
+       :defs
+       (filter (every-pred named? public? (complement foreign-protocol?)))
+       (into {})))
+
+(defn public-macros
+  "Given a namespace return all the public var analysis maps. Analagous to
+  clojure.core/ns-publics but returns var analysis maps not vars.
+
+  Inspired by the ns-publics in cljs.analyzer.api."
+  [env ns]
+  {:pre [(symbol? ns)]}
+  #?(:clj (when (and ns (clojure.core/find-ns ns))
+            (->> (ns-publics ns)
+                 (filter macro?)
+                 (into {})))
+     :cljs (->> (merge
+                 (get-in env [NSES ns :macros])
+                 (get-in env [NSES ns :defs]))
+                (remove (fn [[k v]] (:private v)))
+                (into {}))))
+
+(defn core-vars
+  "Returns a list of cljs.core vars visible to the ns."
+  [env ns]
+  (let [vars (public-vars env 'cljs.core)
+        excludes (:excludes (find-ns env ns))]
+    (apply dissoc vars excludes)))
+
+(defn core-macros
+  "Returns a list of cljs.core macros visible to the ns."
+  [env ns]
+  (let [macros (public-macros env #?(:clj 'cljs.core :cljs 'cljs.core$macros))
+        excludes (:excludes (find-ns env ns))]
+    (apply dissoc macros excludes)))
+
+(defn ns-interns-from-env
+  "Given a namespace return all the var analysis maps. Analagous to
+  clojure.core/ns-interns but returns var analysis maps not vars.
+
+  Directly from cljs.analyzer.api."
+  [env ns]
+  {:pre [(symbol? ns)]}
+  (merge
+   (get-in env [NSES ns :macros])
+   (get-in env [NSES ns :defs])))
+
+(defn sanitize-ns
+  "Add :ns from :name if missing."
+  [m]
+  (-> m
+      (assoc :ns (or (:ns m) (:name m)))
+      (update :ns u/namespace-sym)
+      (update :name u/name-sym)))
+
+(defn var-meta
+  "Return meta for the var, we wrap it in order to support both JVM and
+  self-host."
+  [var]
+  (cond-> {}
+    (map? var) (merge var)
+    (var? var) (-> (merge (meta var))
+                   (update :ns #(cond-> % (u/ns-obj? %) ns-name)))
+    true sanitize-ns
+    #?@(:cljs [true (-> (update :ns u/remove-macros)
+                        (update :name u/remove-macros))])))
+
+(defn ns-meta
+  "Return meta for the var, we wrap it in order to support both JVM and
+  self-host."
+  [var]
+  (cond-> {}
+    (map? var) (merge var)
+    (u/ns-obj? var) (merge {:ns (ns-name var)
+                            :name (ns-name var)})
+    true sanitize-ns
+    #?@(:cljs [true (-> (update :ns u/remove-macros)
+                        (update :name u/remove-macros))])))
+
+(defn find-symbol-meta
+  "Given a namespace-qualified var name, gets the analyzer metadata for that
+  var."
+  [env sym]
+  (let [ns (find-ns env (u/namespace-sym sym))]
+    (some-> (:defs ns)
+            (get (u/name-sym sym))
+            var-meta)))
+
+(defn special-meta
+  "Given a special symbol, gets the analyzer metadata."
+  [_ sym]
+  (when-let [meta #?(:clj (or (get (u/require-and-resolve 'cljs.repl/special-doc-map) sym)
+                              (get (u/require-and-resolve 'cljs.repl/repl-special-doc-map) sym))
+                     :cljs (or (get special/special-doc-map sym)
+                               (get special/repl-special-doc-map sym)))]
+    (merge {:name sym
+            :ns 'cljs.core
+            :special-form true}
+           meta)))
--- a/src/orchard/cljs/analysis.cljc
+++ b/src/orchard/cljs/analysis.cljc
@@ -2,8 +2,7 @@
       :author "Gary Trakhman"
       :added "0.6.0"}
  orchard.cljs.analysis
-  (:require [orchard.misc :as u]
-            #?(:clj [cljs.repl]))
+  (:require [orchard.misc :as u])
   (:refer-clojure :exclude [find-ns find-var all-ns ns-aliases]))
 
 (def NSES :cljs.analyzer/namespaces)
--- a/src/orchard/info.clj
+++ b/src/orchard/info.clj
@@ -1,12 +1,154 @@
 (ns orchard.info
+  "Retrieve the info map from var and symbols."
+  (:refer-clojure :exclude [qualified-symbol?])
   (:require
    [clojure.edn :as edn]
    [clojure.java.io :as io]
    [clojure.java.javadoc :as javadoc]
    [orchard.classpath :as cp]
    [orchard.java :as java]
+   [orchard.cljs.analysis :as cljs-ana]
+   [orchard.cljs.meta :as cljs-meta]
    [orchard.meta :as m]
    [orchard.misc :as u]))
+
+(defn normalize-ns-meta
+  "Normalize cljs namespace metadata to look like a clj."
+  [meta]
+  (merge (select-keys meta [:doc :author])
+         {:file (-> meta :defs first second :file)
+          :line 1
+          :name (:name meta)
+          :ns (:name meta)}))
+
+(defn qualify-sym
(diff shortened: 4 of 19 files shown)


arichiardi:
@bbatsov in porting I noticed that the namespace in Clojure is returned as a `Namespace` object instead of a `symbol`...I think we should normalize it and because ClojureScript does not reify namespaces the only option is to use symbols.

Now...I guess it would be breaking at some point up the `cider-nrepl` chain but I will try to test and document it.

What do you think?

bbatsov:
I guess that won't be a big deal, as most people use only the nREPL interface exposed by cider-nrepl.

arichiardi:
@bbatsov so I ported it finally. This means that after I fix the CI, this is good to go. However, the API changed (will write how in the README) so I think this should go in together with changes is `cider-nrepl`.

Another note is that this PR does not change behaviors, for instance I have not included specs in the Cljs meta at the moment. That was my original plan and so it might be actually a good idea to review this but keep adding commits to this branch, test everything and then bump all the thing!

What do you think?

cichli:
IMHO it is better to keep the CLJS support as a separate library.

* Some of the differences between Clojure and ClojureScript are major, some are subtle, and trying to accommodate all of these differences in one place can get painful.
* Clojure and ClojureScript evolve at different rates. By separating the libs we follow these rates of change separately.
* The self-host stuff is just not relevant to the CLJ parts of this library (N.B. I have simplified the self-host build and CI upstream, so if we do merge this, I would appreciate those changes being merged too).
* With this PR the git history of the existing library is lost. This is valuable information when debugging/changing the code, and it's a shame to lose it unnecessarily.

I think it's more reasonable to just reimplement the spec stuff in `cljs-tooling`. We could make a shared library for any truly common code, but we should be careful about this, because if at some point in the future either CLJ or CLJS diverges in behaviour from what the shared library expects, we get in a mess.

arichiardi:
> IMHO it is better to keep the CLJS support as a separate library.

I thought about this, but there are two major disadvantages:

* Cljs tooling is never growing with the Clj tooling
* Nobody is willing to maintain yet another library

Note that the behavior is still separate in the code, the `:dialect` key controls which function will be called, but now we have better testing for `info`. Also, things like cleaning up and normalization of the generated return map now can be extracted and I really saw in the code that there are many places where there are too many things going on at the same time.

> Clojure and ClojureScript evolve at different rates. By separating the libs we follow these rates of change separately.

Semantically they are the same language so they should have the same API for retrieving info and even completion I maintain.

> The self-host stuff is just not relevant to the CLJ parts of this library (N.B. I have simplified the self-host build and CI upstream, so if we do merge this, I would appreciate those changes being merged too).

I don't have a clear vision here of why self-host is not relevant, given we have conditional readers, which is a Clojure feature, I don't see why self-host should not be considered. In particular, for many of the info stuff returned there could be a self-host server that actually returns it. Having said that, I think in this first iteration I might just get the Cljs part in, I don't know yet, depends on how much time I have.

> With this PR the git history of the existing library is lost. This is valuable information when debugging/changing the code, and it's a shame to lose it unnecessarily.

What do you mean by this? This is a PR and as every commit you have git history, am I missing something?

arichiardi:
Oh I see you where talking about the `.clj` to `.cljc` rename. I could do it in stages for sure. It would be probably cleaner. So no self-host for now. How does it sound?

cichli:
>Nobody is willing to maintain yet another library

I am willing to do that.

> Also, things like cleaning up and normalization of the generated return map now can be extracted and I really saw in the code that there are many places where there are too many things going on at the same time.

This sounds like a good change but it's not clear what you mean looking at the diff in Github. I will look locally with `--find-renames` later.

>Semantically they are the same language so they should have the same API for retrieving info and even completion I maintain.

They are different in [quite a few ways](https://clojurescript.org/about/differences). We can still maintain equivalent APIs in two libraries, right?

>I don't have a clear vision here of why self-host is not relevant

For example, the classpath and javadoc features. 

> What do you mean by this? This is a PR and as every commit you have git history, am I missing something?

Your PR doesn't contain the commits from the other library in e.g. the `analysis` namespace. Tools like `git blame` and `git bisect` don't work as well after code churn like this. AFAIK it is possible to carry the history across to another repo (I believe this was done when some parts of `cider-nrepl` were originally moved into this library). It's not a dealbreaker but no need to avoid it if we can?

arichiardi:
> They are different in quite a few ways. We can still maintain equivalent APIs in two libraries, right?

Why would you want that and what benefits does it bring? Then every time there is a change in clj you have to replicate it in `cljs-tooling` and I have already seen that this does not happen. This is also publicly evident from the feature disparity between the two. Your point would be valid but unfortunately syncing things never happens in reality.

> For example, the classpath and javadoc features.

I contributed to lumo for quite a while and I feel that Cljs imposes the classpath constraint on any self-host implementation. That is why lumo has a `lumo.classpath` layer to replicate what the JVM version has.
I don't think javadoc is relevant as a feature at all here because it does not involve meta reading and of course would not appear in a JS runtime

> I am willing to do that.

While I really trust that you would maintain the code, there are just so much more benefits in having a central place where all the code goes. For example orchard is used by Calva so, which now has to include two libraries and do chores, bump, all that. With one solution Orchard would become *the only thing* you need in order to have feature parity with Cider. This is a dream for now but this is a good start.

arichiardi:
> library in e.g. the analysis namespace. Tools like git blame and git bisect don't work as well after code churn like this. AFAIK it is possible to carry the history across to another repo (I believe this was done when some parts of cider-nrepl were originally moved into this library).

I see this as the weakest of the cons you have made so far so I would really ignore it for this discussion if you agree to do so. In my experience testing things properly has always been more important than git history but maybe I just got lucky :wink:

arichiardi:
@bbatsov this now is all green and does not include self-host and clj->cljc renaming in `src`. Just some in the tests.

bbatsov:
I’m on a vacation with a computer for a while. I’ll leave the review part to @cichli and @SevereOverfl0w in my absence. :-)

arichiardi:
@cichli and @SevereOverfl0w and @bbatsov 

My next task, and I would like to start today maybe, would be to copy the leftover "business logic" from the `info` middleware to the `orchard` `info` function. This will make the `info` function in `orchard` self-contained, which is great, easier to test and more portable (one could use `orchard` without the `nrepl` protocol for instance - say for a Language Server Protocol implementation).

Having said that, it's been a couple of weeks and I did not see any further activity on this PR, so I am going to ask the very controversial question "is anybody interested in seeing this in?". I know for instance @cichli is against it, but I hoped I could convince him otherwise :smile: 

If not, well I will just working on it :+1:

bbatsov:
I won’t be around until early March. In the mean time you’ll have to figure out ways to make progress without me.

arichiardi:
The last commit containing the `boot` tweak + test applied kinda completes the port.

Now I will start trying to figure out how to assoc the spec info in ClojureScript as well so that we would finally have feature parity.
I am kind of proud of the tests, which are way more comprehensive than before.

arichiardi:
@cichli and @SevereOverfl0w and @bbatsov

In order to test this patch properly I was thinking of booking up Circle. Maybe it would be a good thing if somebody could help with enabling it for this repo...if I am made collaborator it would be even easier for me to work and maintain that as well.

I think only a few folks now are aware of the info part of the code base and I would like to volunteer myself to maintain and merge PR in this repo.

This one would be the the first. I have a fork of cider-nrepl now that works against this, which does not change any functionality at the moment, just to make sure we don't break things.

cichli:
FYI: I'm not 100% against this. I think you're right that having a single library to depend upon for feature parity with CIDER is compelling. But this is a large PR, and I might not have time to review it in its entirety for the next few days at least :-).

>In order to test this patch properly I was thinking of hooking up Circle.

Moving to Circle would be very welcome; I know @shen-tian was intending to look at this, so it might be worth coordinating with them on that. @bbatsov will have to enable it for this repo.

Could we also merge in the build changes from upstream? It's [all in the Makefile now](https://github.com/clojure-emacs/cljs-tooling/blob/master/Makefile#L8-L18). See clojure-emacs/cljs-tooling@fa9998fa8c3d478144613ae18ef6f48eb017ffc6.

>I am kind of proud of the tests, which are way more comprehensive than before.

Fantastic work! Not strictly necessary in this PR but it would be really nice if we could remove any dependencies on core.async, mount, etc. from the CLJS tests. In hindsight, the convenience this offers isn't worth how painful it makes testing multiple CLJS versions. I think any examples we need for testing, we should define ourselves.

> I have a fork of cider-nrepl now that works against this, which does not change any functionality at the moment just to be clear, I am using it to make sure I don't break things.

Great stuff – could you link to that fork too? I will clone both and give it a try when I get round to reviewing this in depth.

arichiardi:
> Could we also merge in the build changes from upstream? It's [all in the Makefile now](https://github.com/clojure-emacs/cljs-tooling/blob/master/Makefile#L8-L18). See [clojure-emacs/cljs-tooling@fa9998f](https://github.com/clojure-emacs/cljs-tooling/commit/fa9998fa8c3d478144613ae18ef6f48eb017ffc6).

Will definitely do this, thanks for the PR over there ;)

> Fantastic work! Not strictly necessary in this PR but it would be really nice if we could remove any dependencies on core.async, mount, etc. from the CLJS tests. In hindsight, the convenience this offers isn't worth how painful it makes testing multiple CLJS versions. I think any examples we need for testing, we should define ourselves.

Ok yes this makes total sense, will queue that up :smile:

arichiardi:
Oh cool, just found a problem to fix - leaving the op here for myself:

```
{:nrepl.middleware.print/stream? 1
 :nrepl.middleware.print/print 'cider.nrepl.pprint/zprint-pprint
 :nrepl.middleware.print/quota 1048576
 :nrepl.middleware.print/options {:width 79}
 :op stacktrace
 :session 18c68a74-dc83-4939-82c1-ac3e9e32fe6d
 :id 160}
```

shen-tian:
@arichiardi: I've got a PR in piggieback for CI on CircleCI in case it's useful: https://github.com/nrepl/piggieback/pull/102/files

I imagine there's gonna be some common tasks across nrepl/clojue-emacs to be extracted at some stage.

Is there anything you are looking for, in particular, in the build/test process?

arichiardi:
@shen-tian that looks great, I will copy and tweak :smile: 

Hopefully it will not differ that much from yours.

bbatsov:
The CircleCI config just got in. I'll try to find some time to review this PR (finally) in the next week or so.

arichiardi:
Awesome thank you, will rebase.

I am going to write comments around pain points I also would like feedback on.

arichiardi:
I am using this patch in my `cider` at the moment and haven't noticed any problem so far.

src/orchard/info.clj
```diff
@@ -1,35 +1,169 @@
...
   (edn/read-string (slurp (io/resource "see-also.edn"))))
 
-(defn see-also
-  [ns sym]
-  (let [var-key (str ns "/" sym)]
-    (->> (get see-also-data var-key)
-         (filter (comp resolve u/as-sym)))))
+(defn see-also*
+  [{:keys [dialect unqualified-sym] :as m}]
+  (when-not (= :cljs dialect)
```
arichiardi: In another patch I am planning to change `see-also.edn` so that we can support ClojureScript as well.

src/orchard/meta.clj
```diff
@@ -101,6 +104,13 @@
 ;; They just map to a special symbol.
 (def special-sub-symbs '{& fn*, catch try, finally try})
 
+(defn repl-special-meta
```
arichiardi: This function has been added in order to dynamically require either `clojure.repl` or `cljs.repl` depending on what's available on the classpath - `require-and-var` should be able to avoid resolving the var every time and do it only once. I will make sure it is like that. Any comment/feedback on how to better require dynamically vars is greatly appreaciate here.

src/orchard/misc.clj
```diff
@@ -2,11 +2,8 @@
   (:require
    [clojure.string :as str]))
 
-(def ^:const windows-prefix
-  "Windows")
-
 (defn os-windows? []
-  (.startsWith (System/getProperty "os.name") windows-prefix))
+  (.startsWith (System/getProperty "os.name") "Windows"))
```
arichiardi: Saving one var loading, tiny perf improvement :smile:

src/orchard/misc.clj
```diff
@@ -99,3 +96,70 @@
 
 ;; handles vectors
 (prefer-method transform-value clojure.lang.Sequential clojure.lang.Associative)
+
+;; from https://github.com/flatland/useful/blob/develop/src/flatland/useful/experimental.clj#L31
+(defmacro cond-let
```
arichiardi: This will go away as it is not used anymore

src/orchard/misc.clj
```diff
@@ -99,3 +96,70 @@
...
+          str
+          (str/replace #"\$macros" "")
+          symbol))
+
+(defn ns-obj?
+  "Return true if n is a namespace object"
+  [ns]
+  (instance? clojure.lang.Namespace ns))
+
+(defn require-and-var
```
arichiardi: The purpose is to resolve the var in the namespace and return the required var.

I would like to avoid doing this too many times and I think I am actually not achieving this at all here...

project.clj
```diff
@@ -32,15 +33,30 @@
   :profiles {
              ;; Clojure versions matrix
              :provided {:dependencies [[org.clojure/clojure "1.10.0"]]}
-             :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
-             :1.9 {:dependencies [[org.clojure/clojure "1.9.0"]]}
-             :1.10 {:dependencies [[org.clojure/clojure "1.10.0"]]}
+             :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]
+                                  [org.clojure/clojurescript "1.10.520" :scope "provided"]]}
```
arichiardi: @cichli @bbatsov
As you know the Cljs version previously tested was matching the Clojure minor version (`1.8.0` -> `1.8.51`). 

Unfortunately exactly `1.8.51` has a problem, it is so so old that it is not even reading `clojure.core.async` from the classpath when compiling - probably a bug. I tested a couple of `1.9`s on Maven and it seems like the fix was around the `1.9.2XX` series.

I had to make a choice here. In my daily ClojureScript, the most recent version is the one that I use. I personally have never ever even thought of using something ancient like `1.8.51`.

Does it then make sense what I did there - aka support only the latest?

All the tests are green this way but I am going to wait for feedback here.

bbatsov: I wonder if anyone uses ClojureScript other than the most recent version. Not a big deal.

bbatsov:
Why did you change the format of the see-also.edn exactly? The current format is exactly what we get from an export from clojuredocs.org. Changing it would also require writing some task to pull and transform the data from there.

project.clj
```diff
@@ -32,15 +33,30 @@
...
              :master {:repositories [["snapshots"
                                       "https://oss.sonatype.org/content/repositories/snapshots"]]
-                      :dependencies [[org.clojure/clojure "1.11.0-master-SNAPSHOT"]]}
+                      :dependencies [[org.clojure/clojure "1.11.0-master-SNAPSHOT"]
+                                     [org.clojure/clojurescript "1.10.520" :scope "provided"]]}
 
              :sysutils {:plugins [[lein-sysutils "0.2.0"]]}
 
+             ;; DEV tools
+             :dev {:dependencies [[pjstadig/humane-test-output "0.9.0"]
```
bbatsov: What do you use those for?

arichiardi:
> Why did you change the format of the see-also.edn exactly? The current format is exactly what we get from an export from clojuredocs.org. Changing it would also require writing some task to pull and transform the data from there.

Oh well didn't exactly know that, it is not written anywhere :smile: 

It looked like we were doing extra work for the conversion to string->symbol and viceversa all the time. I will bring it back as it was.

project.clj
arichiardi: to improve the test output - this gives nice diffs - https://github.com/pjstadig/humane-test-output#clojure

arichiardi: Can of course take it out if you don't like it...it is a `:dev` dep anyways...

src/orchard/classpath.clj
```diff
@@ -51,3 +53,21 @@
    (classpath-jarfiles (classpath)))
   ([path]
    (map #(JarFile. ^File %) (filter cp/jar-file? path))))
+
+(defn classpath-file-relative-path
```
bbatsov: Don't we have a patch about this in the `classpath` function already?

bbatsov:
> Not strictly necessary in this PR but it would be really nice if we could remove any dependencies on core.async, mount, etc. from the CLJS tests. In hindsight, the convenience this offers isn't worth how painful it makes testing multiple CLJS versions. I think any examples we need for testing, we should define ourselves.

👍 The simpler the setup is, the better.

bbatsov:
The README should also be updated to reflect to ClojureScript support.

src/orchard/classpath.clj
arichiardi: I will double check but this was in the `info` middleware in `cider-nrepl` and it's been ported over here

src/orchard/cljs/analysis.cljc
```diff
@@ -0,0 +1,203 @@
+(ns orchard.cljs.analysis
```
bbatsov: Please, add some descriptions (as informative as possible) to all new namespaces.
They should also have `{:added "0.6"}` in their metadata. I plan to cut one release before merging this.

arichiardi: Ok will do

arichiardi:
Will update the README

src/orchard/cljs/analysis.cljc
```diff
@@ -0,0 +1,203 @@
...
+  (:require [orchard.misc :as u]
+            #?(:cljs [orchard.cljs.special :as special]))
+  (:refer-clojure :exclude [find-ns find-var all-ns ns-aliases]))
+
+(def NSES :cljs.analyzer/namespaces)
+
+(defn all-ns
```
bbatsov: I guess you should also implement `orchard.query` based on this.

arichiardi: yeah that's what I was asking yesterday, not really familiar with that namespace and it deserves a new PR I guess

src/orchard/meta.clj
arichiardi: @bbatsov @cichli I am intersted in your opinion on this, in particular how to make it faster

src/orchard/misc.clj
bbatsov: Won't `require-and-resolve` be a better name for this? Or `requiring-resolve`?

arichiardi: sounds like a better name yeah

src/orchard/classpath.clj
arichiardi: Not sure about what is going on in this function actually, so I would rather leave it here for now cause it can break functionality to change it. I can open a separate issue when merged to track it.

arichiardi:
I added two new commits for trying some optimizations and cleaned up the PR.

I will as usually try this locally.

bbatsov:
Can you rebase on top of `master`?

arichiardi:
@bbatsov Done

bbatsov:
Small update - I'll merge this after @jeffvalk is ready with #20, as I want to cut one version of Orchard before we introduce such big changes.

arichiardi:
@bbatsov Thank you for the update, really appreciate

arichiardi:
Rebasing this boy to master

bbatsov:
I see lots of conflicts now.

arichiardi:
Rebase took me more time, will ping you when done

arichiardi:
@bbatsov @cichli 

This has been rebased and as a bonus I removed the dependencies on `core.async` and `mount` for the tests.

bbatsov:
I see you brought back the boot hacks here. I'm really on the fence about them. After @jeffvalk removed them I thought it might be better for clients to decide what to do about boot projects. I don't know. I find it funny the code is jumping back and forth, though. :-)

arichiardi:
No well I didn't know it got removed in the first place 😄

This is a straight port from cider-nrepl so if removed it from there I remove it here too I guess

arichiardi:
Oh I see this https://github.com/clojure-emacs/cider-nrepl/commit/22bd2c65ef61f5a1b1d6d27d07481e56a4298a24

Cool I will take it away

arichiardi:
No wait that's restoring it 😄😄 Now I am confused.

bbatsov:
My point is - do we know for Orchard to know about boot or not? Generally both approaches make some sense. If only Boot wasn't messing with the classpath to begin with...

arichiardi:
IMHO yes because as orchard user I would not want to replicate the boot hack...that is why I like this idea of self-contained info op...

jeffvalk:
> If only Boot wasn't messing with the classpath to begin with...

I'm not sure what Boot does to the classpath, but whatever it is, it's either visible to a classloader that the user sees or it isn't. Shouldn't querying the classpath be sufficient?

arichiardi:
Not sure about the details either - take this with a grain of salt - I think there is a file path transformation going on because boot caches things in a folder and so we need to reconcile them.

SevereOverfl0w:
> My point is - do we know for Orchard to know about boot or not? Generally both approaches make some sense. If only Boot wasn't messing with the classpath to begin with...

In my opinion: Boot should be factored into orchard, yes.

> I'm not sure what Boot does to the classpath, but whatever it is, it's either visible to a classloader that the user sees or it isn't. Shouldn't querying the classpath be sufficient?

Boot copies your source code into another folder, and the other folder is the _real_ classpath. Boot provides access to the original classpath (which includes your original src/ folders) via a system property. If you want to support `info` jumping to source, you _must_ include the boot hacks.

bbatsov:
Well, fair enough. But I think this should be added back in a different PR, so we can restore the old behaviour before cutting Orchard 0.5. And probably we should discuss the extend to which we want to use extend this - whether it's just adjust the paths in info or do we want to tweak `orchard.classpath` as well.

bbatsov:
More about this - https://github.com/boot-clj/boot/issues/249 (and related tickets)

arichiardi:
Just a note though, there it no behavior change from `cider` perspective. Info was and stays the same. Agree maybe the classpass functionality should be tweaked but it is definitely a bigger patch.

Anyways, open to any decision

arichiardi:
If I understand things right, this could go in as is and then we can follow up with a patch that modifies classpath and info to use classpath?

bbatsov:
Yeah, that's an option. I'm just a bit frustrated that we've been moving some code back and forth between Orchard and cider-nrepl. Generally I want to limit the number of user facing changes, so I'd be good if we settled on the API in version 0.5.

src/orchard/misc.clj
```diff
@@ -156,3 +156,17 @@
   "Return true if n is a namespace object"
   [ns]
   (instance? clojure.lang.Namespace ns))
+
+;; Drop this in favor of clojure.core/requiring-resolve at some point?
```
bbatsov: I think they made it private.

src/orchard/meta.clj
```diff
@@ -104,13 +104,17 @@
...
-(defn repl-special-meta
+(defn repl-special-meta*
+  "Return the REPL specials info."
   [sym]
   (or (when-let [f (u/require-and-resolve 'clojure.repl/special-doc)]
         (f sym))
       (when-let [f (u/require-and-resolve 'cljs.repl/special-doc)]
         (f sym))))
 
+(def ^{:doc "Return the REPL specials info. Memoized."} repl-special-meta
```
bbatsov: Why don't you use a normal docstring here?

src/orchard/meta.clj
```diff
@@ -104,13 +104,17 @@
 ;; They just map to a special symbol.
 (def special-sub-symbs '{& fn*, catch try, finally try})
 
-(defn repl-special-meta
+(defn repl-special-meta*
```
bbatsov: Why `repl-special-meta*` instead of simply `special-meta*`?

src/orchard/cljs/analysis.cljc
```diff
@@ -0,0 +1,206 @@
+(ns ^{:doc "ClojureScript analysis functions."
+      :author "Gary Trakhman"
+      :added "0.6.0"}
+ orchard.cljs.analysis
+  (:require [orchard.misc :as u]
+            #?(:clj [cljs.repl]))
+  (:refer-clojure :exclude [find-ns find-var all-ns ns-aliases]))
+
+(def NSES :cljs.analyzer/namespaces)
```
bbatsov: I guess that's meant to be private, right? In general I don't see much point in mapping a constant to a keyword.

src/orchard/info.clj
```diff
@@ -1,13 +1,155 @@
...
+     ;; var in cljs.core
+     (some->> (get (cljs-ana/core-vars env context-ns) sym)
+              (cljs-ana/var-meta)
+              (cljs-meta/normalize-var-meta))
+     ;; macro in cljs.core
+     (some->> (cljs-meta/scoped-macro-meta env sym 'cljs.core)
+              (cljs-meta/normalize-macro-meta)))))
+
+(defn info
+  [params]
```
bbatsov: Switching to a map here is a breaking change. Probably this function should have two signatures to keep backward compatibility (one taking a symbol and ns, and the other taking the map). Not to mention this key function should have a docstring the size of a novel. :D

bbatsov: Maybe even a third signature - sym, ns and dialect.

bbatsov:
@arichiardi I left a bit more feedback (mostly concerns about the breaking change for `info`). I think we have been dragging this for long enough, so in light of everything I'll just merge and release it as part of Orchard 0.5 when you're ready with the changes.

src/orchard/cljs/analysis.cljc
arichiardi: Yes, you are right, it was like that but I can change it

src/orchard/meta.clj
arichiardi: Well they are two different things in theory but let me double check if I am doing the right thing

src/orchard/info.clj
arichiardi: How many `orchard` users do we have? 

I kind of think this is a breaking change worth having but if you think it's too much I can change it of course

bbatsov: > How many orchard users do we have?

No idea. Probably not many, but I don't think it makes sense to break the API if you can simply add a couple of extra arities to the function.

arichiardi: Ok will change it

src/orchard/misc.clj
alexander-yakushev: `requring-resolve` is public. They made `serialized-require` private which ensures requiring is done under a lock.

arichiardi:
Addressed almost everything but still need to complete one thing.

arichiardi:
Ok done, this addresses the latest set of feedback, it's again ready to go ;)

project.clj
```diff
@@ -4,7 +4,8 @@
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :scm {:name "git" :url "https://github.com/clojure-emacs/orchard"}
-  :dependencies [[org.tcrawley/dynapath "1.0.0"]]
+  :dependencies [[org.tcrawley/dynapath "1.0.0"]
+                 [org.clojure/clojurescript "1.10.520"]]
```
bbatsov: Shouldn't this be a provided dep instead of a hard dep?

bbatsov:
Apart from the hard dep on ClojureScript, which should be made provided I think we're good.

arichiardi:
I don't know, shouldn't it be an hard dep? The lib does not work without it and you can always override it in your own project.clj.

I never found a reason for provided in general and I remember some boot folk saying that it is even bad (but can't recall now).

Having said that, maybe there is a reason to use it in which case I will change it

bbatsov:
OK, fair enough. 

Thanks for the hard work and for your persistence to see this PR through! 🙇 
I'll cut a new beta right away, so you can proceed integrating this into `cider-nrepl`.

arichiardi:
Thank you!

This is super great and can't wait for it to test it out (and fix the bugs 😁😁).

The other part of this is the PR to compliment after which we can say bye to `cljs-tooling` 😍🤠

arichiardi:
TODOs here (for myself mainly):

* Get rid of the `boot` leftover and make sure the boot jump to def works #52 
* Possibly merge https://github.com/clojure-emacs/cider-nrepl/pull/568

PR #38, opened 2019-01-02
Add eval and def features to inspector

alexander-yakushev:
Inspired by REBL, these are two features that would subsequently be available in CIDER inspector:

- `def` the currently inspected value as a var.
- prompt the user for an expression and eval it in the context where `v` is bound to the currently inspected value. Inspect the result just to be consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+* [#38](https://github.com/clojure-emacs/orchard/pull/38): [Inspector] Add eval
+  and def features to inspector.
+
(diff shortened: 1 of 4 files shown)


bbatsov:
That's pretty cool! 

Don't forget to mention the changes in the changelog.

bbatsov:
Btw, can you also take a look a this failure on JDK 11? Seems some class got removed/renamed.

alexander-yakushev:
This looks like a separate problem. Do you want to handle it in this PR?

bbatsov:
Thanks!

bbatsov:
The latest cider-nrepl already includes your changes.

PR #39, opened 2019-01-10
Catch LinkageError when calling Class/forName

cichli:
Fixes clojure-emacs/cider#2562. See https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.2 for a far more exhaustive discussion of what is actually going on during class loading/initialisation, and why we might need to guard against these errors when calling `Class/forName` – especially when native libraries are involved.


bbatsov:
Thanks!

PR #42, opened 2019-01-25
Namespace alias support for `spec-list` and `spec-form`

tatut:
Added support for resolving namespace aliased keywords for the current namespace.

Changes:

- `spec-list` will return "::some-ns-alias/some-keyword" completions in addition to the fully qualified names for specs in namespaces that have an alias in the given ns.
- `spec-list` will return "::some-keyword" completions in addition to the fully qualified names for specs in the given ns.
- `spec-form` will expand both ns aliases before resolving the spec
- added tests for ns alias functionality

I will add a separate PR for CIDER to add `(cider-current-ns)` for `spec-list` and `spec-form` requests in `cider-client.el` if this is accepted.

### DIFF
--- a/test/orchard/spec_test.clj
+++ b/test/orchard/spec_test.clj
@@ -21,30 +21,34 @@
              :ret (clojure.core/fn [%] (clojure.core/> (:start %) (:end %)))
              :fn nil)))))
 
-(deftest spec-is-found-by-ns-alias
-  (testing "current ns keyword"
-    (testing "spec-list finds current ns keyword"
-      (eval '(clojure.spec.alpha/def ::foo string?))
-      (let [specs (into #{}
-                        (spec/spec-list "" "orchard.spec-test"))]
-        (is (specs "::foo") "Spec is found with current ns")
-        (is (specs ":orchard.spec-test/foo") "Spec is found with fully qualified name")))
+(def spec-available? (or (resolve (symbol "clojure.spec.alpha" "get-spec"))
+                         (resolve (symbol "clojure.spec" "get-spec"))))
 
-    (testing "spec-form finds current ns keyword"
-      (let [spec1 (spec/spec-form "::foo" "orchard.spec-test")
-            spec2 (spec/spec-form ":orchard.spec-test/foo" "orchard.spec-test")]
-        (is (= "clojure.core/string?" spec1 spec2) "Both return the same correct spec"))))
+(when spec-available?
+  (deftest spec-is-found-by-ns-alias
+    (testing "current ns keyword"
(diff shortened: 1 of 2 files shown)


bbatsov:
Sorry for the slow turnaround here! (I was away from a computer for quite a while) Can you rebase on `master`, so I can see the tests will pass with new CI?

The changes look good to me, but before merging this I wanted to ping @alexander-yakushev about this. It might make sense to have similar behavior for keyword in compliment for consistency's sake.

test/orchard/spec_test.clj
```diff
@@ -20,3 +20,35 @@
                     (clojure.core/fn [%] (clojure.core/< (:start %) (:end %))))
              :ret (clojure.core/fn [%] (clojure.core/> (:start %) (:end %)))
              :fn nil)))))
+
+(def spec-available? (or (resolve (symbol "clojure.spec.alpha" "get-spec"))
```
bbatsov: Now I think there's also alpha2.

alexander-yakushev:
@bbatsov I honestly didn't quite get what is this all about (I don't use spec much, let alone in CIDER). Could you please describe what sort of behavior would you expect from Compliment regarding this?

bbatsov:
I think the idea is provide completion candidates with non-expanded keywords as well as those with expanded ones, when doing something ::some-keyword and :some.ns/keyword, or ::ns-alias/keyword and :real.ns/keyword. Frankly, I rarely use namespaced keywords, so I'm not 100% certain how useful this would be, but it kind of makes sense to me.

tatut:
Yes, the point is to be able to lookup keywords with ns aliases.

I use them alot, so that I will have a require like
```(:require [mything.foo-domain :as foo])``
` 

And later use keywords like `::foo/bar`. This will (with the upcoming CIDER) change allow you to lookup keyword under point and it will correctly understand that it means `:mything.foo-domain/bar` and get the correct spec for it.

bbatsov:
Well, apart from my small remark about spec2, I'm ready to merge this.

tatut:
I'll touch this up this evening and make sure the tests still pass.

alexander-yakushev:
Ah. Compliment already does that, if I remember correctly.

test/orchard/spec_test.clj
tatut: As far as I can tell, spec-alpha2 isn't even at pre-release stage and subject to heavy changes... I don't think we should use that API yet.

tatut:
I think this should be ready for merge.

bbatsov:
Thanks for working on this!

PR #44, opened 2019-03-15
Migrate to CircleCI

shen-tian:
Each project has its own challenges... Work in progress, but so far:

- One of the tests makes a false assumption about CI setup (that the project would sit in a dir containing the string `orchard`).  I want to clean this fix up a bit first.
- Looks like the Cloverage tests hasn't passed for a while, here and on Travis. 
- The Circle CI script is exactly the same as Piggieback's, so good sign for reused in the orchard!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,190 @@
+version: 2.1
+
+######################################################################
+#
+# Start of general purpose config. These can plausibly go into orbs
+#
+######################################################################
+
+# Default settings for executors
+
+defaults: &defaults
+  working_directory: ~/repo
+  environment:
+    LEIN_ROOT: "true"   # we intended to run lein as root
+    JVM_OPTS: -Xmx3200m # limit the maximum heap size to prevent out of memory errors
+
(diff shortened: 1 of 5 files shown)


bbatsov:
Seems some code is not properly formatted and that's failing the old build. ;-)

bbatsov:
> Looks like the Cloverage tests hasn't passed for a while, here and on Travis.

There are so often problems with this I wonder why we keep bothering with it.

shen-tian:
I've commented out the cloverage job for now, as it's probably contributing to the broken window effect more than helping? 

Note: the automatic doc gen job wouldn't be running anymore. Do we want to bring that online first?

https://circleci.com/gh/shen-tian/orchard/tree/circleci

bbatsov:
Looks good to me!

> Note: the automatic doc gen/deploy jobs will still be on travis. We can revisit those later?

Yeah, that's fine by me. The doc stuff is not needed anymore, as it predates the existence of cljdoc. The deployment stuff is nice, but it's not that important so we can even drop it completely if it's hard to port it.

PR #46, opened 2019-03-29
Raw object inspector improvements

alexander-yakushev:
1. When inspecting objects that don't have special render handling, separate static and non-static fields. Also, sort fields by name. Looks like this:
image

2. Walk the class hierarchy to display not only immediate object's fields, but also fields inherited from superclasses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### New features
 
 * [#42](https://github.com/clojure-emacs/orchard/pull/42): Add namespace alias support for `spec-list` and `spec-form`.
+* [#46](https://github.com/clojure-emacs/orchard/pull/46): [Inspector] Separate static from non-static fields when rendering raw objects.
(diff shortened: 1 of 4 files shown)


bbatsov:
Looks pretty cool! Seems there's some small issue with the formatting, the code itself looks good to me.

Also - don't forget to update the changelog.

alexander-yakushev:
It looks like cljfmt doesn't understand proper `letfn` indentation.

alexander-yakushev:
Should I rewrite it using regular let or are you willing to spend potentially A LOT of time to investigate? :)

bbatsov:
Probably you can try teaching cljfmt here https://github.com/clojure-emacs/orchard/blob/master/project.clj#L58, but I'd be fine with using a regular `let`.

alexander-yakushev:
Turned out much easier than I thought.

alexander-yakushev:
Thanks, Bozhidar!

bbatsov:
I've also pushed a new snapshot of cider-nrepl.

PR #47, opened 2019-04-04
[java] Cache class-info for editable Java classes

alexander-yakushev:
Not adding tests since caching already hasn't been tested.


bbatsov:
Nicely done!

PR #55, opened 2019-06-07
[Fix #52] Use boot-hack-classloader in file-info

### LINKED ISSUE #52: Jump-to-definition with Boot jumps to temporary files
When using the latest cider-nrepl with the latest Orchard, and after recompiling namespaces with `tools.namespace`, pressing `M-.` on a function name from one of the recompiled namespaces jumps not to the actual file, but to the "temporary" cache file that Boot creates in `~/.boot/tmp/...`.

Looks like the commit that broke it is  https://github.com/clojure-emacs/orchard/commit/b83826f42604f5f9bf59e25a535ef9d3ddb3c5db

A bit more context: 
image

arichiardi:
This patch brings a couple more boot hacks that were previously baked in the
cider-nrepl resource middleware into orchard. This enables the `info/file-info`
method to correctly resolve file resources.

- [X] All tests are passing

### DIFF
--- a/src/orchard/classpath.clj
+++ b/src/orchard/classpath.clj
@@ -2,6 +2,7 @@
   "Classpath access and modification"
   (:require
    [clojure.java.io :as io]
+   [clojure.string :as str]
    [dynapath.util :as dp]
    [orchard.misc :as u])
   (:import
@@ -87,6 +88,28 @@
            (map #(.getPath (.relativize (.toURI url) (.toURI ^File %))))))))
 
 ;;; Boot's hack - previously part of cider-nrepl
+
+(defn boot-classloader
+  "Creates a class-loader that knows original source files paths in Boot project."
+  []
+  (let [class-path (System/getProperty "fake.class.path")
+        dir-separator (System/getProperty "file.separator")
+        paths (str/split class-path (re-pattern (System/getProperty "path.separator")))
+        urls (map
+              (fn [path]
+                (let [url (if (re-find #".jar$" path)
+                            (str "file:" path)
+                            (str "file:" path dir-separator))]
+                  (new java.net.URL url)))
+              paths)]
+    ;; TODO: Figure out how to add the JDK sources here
+    (new java.net.URLClassLoader (into-array java.net.URL urls))))
+
+(defn boot-aware-classloader
+  []
+  (if (u/boot-project?)
+    (boot-classloader)
+    (context-classloader)))
 
 (defn classpath-file-relative-path
   "Boot stores files in a temporary directory & ClojureScript stores
--- a/src/orchard/info.clj
+++ b/src/orchard/info.clj
@@ -201,7 +201,7 @@ resolved (real) namespace and name here"}
   (java/member-info class member))
 
 (defn- resource-full-path [relative-path]
-  (io/resource relative-path (cp/context-classloader)))
+  (io/resource relative-path (cp/boot-aware-classloader)))
 
 (defn resource-path
   "If it's a resource, return a tuple of the relative path and the full resource path."


src/orchard/classpath.clj
```diff
@@ -88,6 +89,28 @@
...
+              (fn [path]
+                (let [url (if (re-find #".jar$" path)
+                            (str "file:" path)
+                            (str "file:" path dir-separator))]
+                  (new java.net.URL url)))
+              paths)]
+    ;; TODO: Figure out how to add the JDK sources here
+    (new java.net.URLClassLoader (into-array java.net.URL urls))))
+
+(defn boot-hack-classloader
```
bbatsov: The name here is a bit misleading, as this might be a boot-classloader or a regular one depending on the project type.

arichiardi: Right, `project-based-classloader`?

You can also see that we have similar stuff at the top of the same file, but I didn't dare replacing it - you are right that is not conditional...

arichiardi: Already turned off the today, preparing for sleeping, will change tomorrow

src/orchard/classpath.clj
```diff
@@ -88,6 +89,28 @@
...
+  []
+  (if (u/boot-project?)
+    (boot-classloader)
+    (.getContextClassLoader (Thread/currentThread))))
```
bbatsov: There's a `context-classloader` helper that does this. I wonder if we shouldn't modify it instead, although I'm not certain if we shouldn't modify it instead.

arichiardi: Maybe yeah I can call that function to avoid future confusion

src/orchard/classpath.clj
bbatsov: Maybe `boot-aware-classloader` or `resource-classloader`?

arichiardi: I like the former

bbatsov:
I think it might also be a good idea to move the project resource lookup into Orchard, so that the conversion of resource paths would be transparent to the clients.

src/orchard/classpath.clj
arichiardi: I was also a bit scared about that TODO 😀

arichiardi:
Agree, will do it in subsequent PRs if you don't mind 😉

bbatsov:
Yeah, that's fine. Just sharing some general thoughts. In the mean time I'll also ping @jeffvalk to review this one as well. 

My other small reservation is the lack of a regression test for this Boot-mess, but I don't think we can easily come up with something meaningful there.

jeffvalk:
> My other small reservation is the lack of a regression test for this Boot-mess, but I don't think we can easily come up with something meaningful there.

Managing Java classpaths has enough complexity by default. Introducing "fake" ones is a recipe for pain.

I know very little about boot, but from it's documentation (which seems quite good), boot's fake classpath is a workaround to support tools that don't handle classpaths properly. Quoting [here][1]:

> This property is a workaround for IDEs and tools that expect the full class path to be determined by the java.class.path property.

Hence, Orchard, in order to be good tooling, is implementing a workaround for a workaround to support bad tooling. This is ugly.

A better solution would be for boot to enable its fake classpath as an option, rather than by default.

Out of curiosity, which IDEs/tools is boot's fake classpath designed to support?

It's worth noting that expecting all resources to be present in "java.class.path" is a broken assumption in Java 9+.

[1]: https://github.com/boot-clj/boot/blob/7709941bcef20de55b940ec18885677df33b7557/boot/core/src/boot/core.clj#L146-L147

bbatsov:
> A better solution would be for boot to enable its fake classpath as an option, rather than by default.
> Out of curiosity, which IDEs/tools is boot's fake classpath designed to support?

Those are great point, but I'm afraid I've never touched boot, so someone else will have to enlighten us. That's why I stay out of boot-related changes, as I simply don't have any idea what's the best course of action there. Maybe @alexander-yakushev or @flyboarder can answer those points you raised.

bbatsov:
I've also noticed this interesting bit of info:

> Also sets the :boot-class-path environment property which is the same as above
  except that the actual class path directories are used instead of the user's
  project directories. This property can be used to configure Java tools that
  would otherwise be looking at java.class.path expecting it to have the full
  class path (the javac task uses it, for example, to configure the Java com-
  piler class).

bbatsov:
I'm once again starting to wonder if the end tools should determine whether they are working in a Boot project and adjust resource paths accordingly based on this. By this I meant that maybe CIDER might do this itself in the Elisp side, and orchard and cider-nrepl might ignore it completely.

jeffvalk:
> I'm once again starting to wonder if the end tools should determine whether they are working in a Boot project and adjust resource paths accordingly based on this. By this I meant that maybe CIDER might do this itself in the Elisp side, and orchard and cider-nrepl might ignore it completely.

This seems the more sensible approach to me.

Boot introducing a workaround for certain tools is reasonable. Letting that workaround propagate into unrelated tools around the ecosystem is pretty undesirable.

arichiardi:
Note that even things like ClojureScript fall exactly into the trap because the `:file` key stored in the compiler state is somehow pointing to the boot cache.

This means the boot hack has already trickled down unfortunately and a probably massive change (ClojureScript compiler understand it's inside a boot project?) would be necessary.

I would argue the whole classloader/resource solution dance is quite a JVM disaster.

bbatsov:
> Note that even things like ClojureScript fall exactly into the trap because the :file key stored in the compiler state is somehow pointing to the boot cache.

Well, that's the real classpath, isn't it? ;-) I assume that's why everything ends up pointing there. My point was that instead of passing some fake classloader around we can just check if we're in a boot project and adjust the file key after the fact. Still ugly, but at least we won't touch the real classloaders.

> I would argue the whole classloader/resource solution dance is quite a JVM disaster.

No argument from me. :D

bbatsov:
Anyways, I guess we'll have to do go forward with this for now for the sake of our users, and think of some proper long-term solution...

jeffvalk:
> Anyways, I guess we'll have to do go forward with this for now for the sake of our users, and think of some proper long-term solution...

Coordinating with the boot maintainers to make boot's fake class path optional seems the right long-term approach. At some point (hopefully soon), orchard version X could remove these hacks, and only support boot versions Y+.

arichiardi:
Ok I will fix the names

arichiardi:
Renamed things :smile:

bbatsov:
Thanks!

> Coordinating with the boot maintainers to make boot's fake class path optional seems the right long-term approach. At some point (hopefully soon), orchard version X could remove these hacks, and only support boot versions Y+.

I'll start some conversation with them.

PR #57, opened 2019-06-10
Fix bug that fails to parse version string without minor version

### LINKED ISSUE #56: Fails to load orchard.java namespace
## Expected behavior

Successfully loading the `orchard.java` namespace.

## Actual behavior

Fails to load the namespace (and which causes `cider-jack-in` to crash).

```
$ clj -Srepro -Sdeps '{:deps {org.clojure/clojure {:mvn/version "1.10.1"} cider/orchard {:mvn/version "0.5.0-beta6"}}}' -e "(require 'orchard.java)"                                                                                                                              [~/private/jise]
Execution error (IllegalArgumentException) at orchard.java/fn (java.clj:65).
No matching field found: getJarFileURL for class sun.net.www.protocol.jrt.JavaRuntimeURLConnection

Full report at:
/var/folders/mr/bf82sldd78g9822yxqgzd12h0000gn/T/clojure-1254707421438656643.edn
$
```

The same thing happens with 0.5.0-beta4 and beta5 as well.

The following is the line that the error message is pointing to:

```clojure
(or (some-> (io/resource "com/sun/javadoc/Doc.class")
            (.. openConnection getJarFileURL))
    (some-> (jdk-find "tools.jar") cp/add-classpath!))
```
And when I dug into those expressions, I saw `(.openConnection ...)` returning `JavaRuntimeURLConnection`, which doesn't have `.getJarFileURL()` method.

```clojure
=> (.openConnection (io/resource "com/sun/javadoc/Doc.class"))
#object[sun.net.www.protocol.jrt.JavaRuntimeURLConnection 0x2d84cb86 "sun.net.www.protocol.jrt.JavaRuntimeURLConnection:jrt:/jdk.javadoc/com/sun/javadoc/Doc.class"]
=> (.getJarFileURL *1)
Execution error (IllegalArgumentException) at user/eval8 (REPL:1).
No matching field found: getJarFileURL for class sun.net.www.protocol.jrt.JavaRuntimeURLConnection
```

The above error happens when I try on OpenJDK 11, but NOT on Oracle JDK 1.8.

### Clojure version

- 1.9.0
- 1.10.0
- 1.10.1

### Java version

```
openjdk version "11" 2018-09-25
OpenJDK Runtime Environment 18.9 (build 11+28)
OpenJDK 64-Bit Server VM 18.9 (build 11+28, mixed mode)
```

### Operating system

macOS Mojave (10.14.5)

athos:
According to [JEP223](http://openjdk.java.net/jeps/223), the minor version and security level parts of a Java version string must be omitted when both numbers are zero.
However, the current implementation of `orchard.misc/java-api-version` doesn't seem to take into account such a case. In fact, on my environment (OpenJDK 11) it fails to parse the version string and returns a wrong version:

```
$ java -version
openjdk version "11" 2018-09-25
OpenJDK Runtime Environment 18.9 (build 11+28)
OpenJDK 64-Bit Server VM 18.9 (build 11+28, mixed mode)
```

```clojure
=> (System/getProperty "java.version")
"11"
=> (require '[orchard.misc :as misc])
nil
=> misc/java-api-version
7
```

This PR fixes the issue and makes `misc/java-api-version` return the correct version:

Oracle JDK 8:
```clojure
=> (System/getProperty "java.version")
"1.8.0_181"
=> (require '[orchard.misc :as misc])
nil
=> misc/java-api-version
8
```

OpenJDK 11:
```clojure
=> (System/getProperty "java.version")
"11"
=> (require '[orchard.misc :as misc])
nil
=> misc/java-api-version
11
```

Also, this should probably fix #56.

### DIFF
--- a/src/orchard/misc.clj
+++ b/src/orchard/misc.clj
@@ -88,16 +88,19 @@
               (last xs)))]
     (apply f (filter identity xs))))
 
-(def java-api-version
+(defn parse-java-version
+  "Parse a Java version string according to JEP 223 and return the appropriate version."
+  [java-ver]
   (try
-    (let [java-ver (System/getProperty "java.version")
-          [major minor _] (str/split java-ver #"\.")
-          major (Integer/parseInt major)
-          minor (Integer/parseInt minor)]
+    (let [[major minor _] (str/split java-ver #"\.")
+          major (Integer/parseInt major)]
       (if (> major 1)
         major
-        (or minor 7)))
+        (Integer/parseInt (or minor "7"))))
     (catch Exception _ 7)))
+
+(def java-api-version
+  (parse-java-version (System/getProperty "java.version")))
 
 (defmulti transform-value "Transform a value for output" type)
 
--- a/test/orchard/misc_test.clj
+++ b/test/orchard/misc_test.clj
@@ -32,6 +32,10 @@
   (is (= (misc/update-keys str {:a :b :c :d :e :f})
          {":a" :b, ":c" :d, ":e" :f})))
 
+(deftest parse-java-version-test
+  (is (= (misc/parse-java-version "1.8.0") 8))
+  (is (= (misc/parse-java-version "11") 11)))
+
 (deftest macros-suffix-add-remove
   (testing "add-ns-macros"
     (is (nil? (misc/add-ns-macros nil)))


bbatsov:
Nice catch!

Can you also add an unit test about this?

src/orchard/misc.clj
```diff
@@ -92,11 +92,10 @@
   (try
     (let [java-ver (System/getProperty "java.version")
           [major minor _] (str/split java-ver #"\.")
-          major (Integer/parseInt major)
-          minor (Integer/parseInt minor)]
+          major (Integer/parseInt major)]
```
bbatsov: Might be good to add some docstring that refers to the JEP you mentioned, so people would understand the logic here better.

athos:
Sure, will do. Just a moment.

athos:
Done. Could you review my code again?

bbatsov:
Looks great! Thanks!

tstirrat15:
I'm seeing something similar on Ubuntu 19.10 right now - similar error and possibly similar cause?

```console
➜  ~ java --version
openjdk 14-ea 2020-03-17
OpenJDK Runtime Environment (build 14-ea+18-Ubuntu-1)
OpenJDK 64-Bit Server VM (build 14-ea+18-Ubuntu-1, mixed mode, sharing)
```
With 
```clojure
user=> (System/getProperty "java.version")
"14-ea"
user=> (require '[orchard.misc :as misc])
nil
user=> misc/java-api-version
7
```
Is it worth opening another issue for this? It looks like these version strings still follow JEP223, but the code that parses it isn't lopping off $OPT when it interprets the version.

PR #58, opened 2019-06-16
Move resource op from cider-nrepl and refactor

arichiardi:
This patch brings in orchard the code necessary for resource resolution and
does a bit of refactoring: javadoc is now in `orchard.java`, where it belongs,
and a new `orchard.java.resource` namespace has been created containing the
actual reusable resource code.

- [X] All tests are passing

### DIFF
--- a/src/orchard/info.clj
+++ b/src/orchard/info.clj
@@ -4,13 +4,13 @@
   (:require
    [clojure.edn :as edn]
    [clojure.java.io :as io]
-   [clojure.java.javadoc :as javadoc]
    [orchard.classpath :as cp]
-   [orchard.java :as java]
    [orchard.cljs.analysis :as cljs-ana]
    [orchard.cljs.meta :as cljs-meta]
+   [orchard.java :as java]
    [orchard.meta :as m]
-   [orchard.misc :as u]))
+   [orchard.misc :as u]
+   [orchard.resource :as resource]))
 
 (defn normalize-ns-meta
   "Normalize cljs namespace metadata to look like a clj."
@@ -200,22 +200,6 @@ resolved (real) namespace and name here"}
   [class member]
   (java/member-info class member))
 
-(defn- resource-full-path [relative-path]
-  (io/resource relative-path (cp/boot-aware-classloader)))
-
-(defn resource-path
-  "If it's a resource, return a tuple of the relative path and the full resource path."
-  [x]
-  (or (if-let [full (resource-full-path x)]
-        [x full])
-      (if-let [[_ relative] (re-find #".*jar!/(.*)" x)]
-        (if-let [full (resource-full-path relative)]
-          [relative full]))
-      ;; handles load-file on jar resources from a cider buffer
-      (if-let [[_ relative] (re-find #".*jar:(.*)" x)]
-        (if-let [full (resource-full-path relative)]
-          [relative full]))))
-
 (defn file-path
   "For a file path, return a URL to the file if it exists and does not
   represent a form evaluated at the REPL."
@@ -228,7 +212,7 @@ resolved (real) namespace and name here"}
 
 (defn file-info
   [path]
-  (let [[resource-relative resource-full] (resource-path path)]
+  (let [[resource-relative resource-full] (resource/resource-path-tuple path)]
     (merge {:file (or (file-path path) resource-full path)}
            ;; Classpath-relative path if possible
            (if resource-relative
@@ -239,25 +223,4 @@ resolved (real) namespace and name here"}
   resources on the classpath; then use online javadoc content for core API
   classes. If no source is available, return the relative path as is."
   [^String path]
-  {:javadoc
-   (or (resource-full-path path)
-       ;; [bug#308] `*remote-javadocs*` is outdated WRT Java
-       ;; 8, so we try our own thing first.
-       (when (re-find #"^(java|javax|jdk|org.omg|org.w3c.dom|org.xml.sax)/" path)
-         (apply str ["https://docs.oracle.com"
-                     (if (>= u/java-api-version 11) "/en/java/javase/" "/javase/")
-                     u/java-api-version
-                     "/docs/api/"
-                     path]))
-       ;; If that didn't work, _then_ we fallback on `*remote-javadocs*`.
-       (some (let [classname (.replaceAll path "/" ".")]
-               (fn [[prefix url]]
-                 (when (.startsWith classname prefix)
-                   (str url path))))
-             @javadoc/*remote-javadocs*)
-       path)})
-
-;; TODO: Seems those were hardcoded here accidentally - we should
-;; probably provide a simple API to register remote JavaDocs.
-(javadoc/add-remote-javadoc "com.amazonaws." "http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/")
-(javadoc/add-remote-javadoc "org.apache.kafka." "https://kafka.apache.org/090/javadoc/index.html?")
+  {:javadoc (java/resolve-javadoc-path path)})
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -3,10 +3,12 @@
   {:author "Jeff Valk"}
   (:require
    [clojure.java.io :as io]
+   [clojure.java.javadoc :as javadoc]
    [clojure.reflect :as r]
    [clojure.string :as str]
    [orchard.classpath :as cp]
-   [orchard.misc :as util])
+   [orchard.misc :as util]
+   [orchard.resource :as resource])
   (:import
    (clojure.lang IPersistentMap)
    (clojure.reflect Constructor Field JavaReflector Method)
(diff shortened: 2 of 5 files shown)


arichiardi:
Found some inconsistency with the way I refactoring but ran out of time. So tagging this as WIP

src/orchard/java/resource.clj
```diff
@@ -0,0 +1,52 @@
...
+
+(defn get-project-resources
+  []
+  (mapcat
+   (fn [directory]
+     (->> directory
+          (file-seq)
+          (filter (memfn isFile))
+          (map (fn [file]
+                 (let [relpath (-> file
```
jeffvalk: FWIW this reimplements [`java.net.URI/relativize`][1]

[1]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URI.html#relativize(java.net.URI)

arichiardi: Is it the exact copy? Is it safe to replace it, meaning, do we have tests against this? :smile: Will investigate

arichiardi:
Removed the WIP status of this fella, should be good now

bbatsov:
@arichiardi Why are there two resource namespaces now - `orchard.java.resource` and `orchard.resource`?

src/orchard/info.clj
```diff
@@ -4,13 +4,13 @@
...
    [clojure.java.io :as io]
-   [clojure.java.javadoc :as javadoc]
    [orchard.classpath :as cp]
    [orchard.java :as java]
    [orchard.cljs.analysis :as cljs-ana]
    [orchard.cljs.meta :as cljs-meta]
    [orchard.meta :as m]
-   [orchard.misc :as u]))
+   [orchard.misc :as u]
+   [orchard.java.resource :as jresource]))
```
bbatsov: This is not sorted alphabetically.

src/orchard/java.clj
```diff
@@ -343,10 +345,37 @@
...
 ;;
 ;; On startup, cache info for the most commonly referenced classes.
 (future
   (doseq [class (->> (ns-imports 'clojure.core)
                      (map #(-> % ^Class val .getName symbol)))]
     (class-info class)))
+
+;; TODO: Seems those were hardcoded here accidentally - we should
+;; probably provide a simple API to register remote JavaDocs.
+(javadoc/add-remote-javadoc "com.amazonaws." "http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/")
```
bbatsov: I think you can safely delete those.

arichiardi:
The former is Java-specific and the latter is the pure impl of the op, very minimal. The former also contains shared code (again all Java-specific).

src/orchard/info.clj
arichiardi: Will change

src/orchard/java.clj
arichiardi: Oh can I? Will do

src/orchard/java/resource.clj
```diff
@@ -0,0 +1,52 @@
+(ns orchard.java.resource
+  "Resolve JVM resource-related paths."
```
bbatsov: It's not really clear what's the difference between the this ns and `orchard.resource`, as both of them are Clojure/Java specific right now.

src/orchard/resource.clj
```diff
@@ -0,0 +1,27 @@
...
+     (.getPath resource))})
+
+(defn resources-list
+  "Return a list of dictionaries containing file and relpath: file is the
+  absolute path to the resource, relpath is the path of the resource
+  relative to the classpath."
+  {:added "0.5.0"}
+  [_]
+  {:resources-list
+   (u/transform-value
```
bbatsov: This is needed only in the context of nREPL responses. Not sure if it makes sense to transform any values for nREPL in Orchard.

src/orchard/java/resource.clj
arichiardi: Ok so maybe there was some overengineering on my part, I can squash them together. It felt tidier to have one ns for the entry point of the op only.

src/orchard/java/resource.clj
```diff
@@ -0,0 +1,52 @@
...
+  {:added "0.5.0"}
+  (:require
+   [clojure.java.io :as io]
+   [orchard.classpath :as cp]
+   [orchard.misc :as u]))
+
+(defn- trim-leading-separator
```
bbatsov: It'd be nice if this function (and all the others) had some docstrings, even though they are relatively simple.

src/orchard/resource.clj
arichiardi: Yeah well I never understood this transform really. It is kind of called in a lot of places and if that's the case (presentation only) probably it should not live in orchard but in cider-nrepl?

I will check what it does

bbatsov: I don't remember what exactly it did, but it was just converting values to types that bencode could represent. I think I copied it accidentally to Orchard and then forgot to move it back.

src/orchard/java/resource.clj
bbatsov: I'm fine with a separation if it's clear what the purpose is - e.g. if the other namespace is something we intend to use to support CLR down the road. Right now they just seemed confusing. 

There's also the fact that we should be designing an optimal API here for general use, not just reflect how things were in `cider-nrepl`. It's fine for the op to have to call a couple of functions instead of one if that would result in a better Orchard API in general.

src/orchard/resource.clj
arichiardi: Oh ok good to know, good target for another PR

bbatsov:
@arichiardi ping :-)

src/orchard/java/resource.clj
arichiardi: I will bring everything in `orchard.resource` for now but I still think we need a better API for this.

arichiardi:
Thinking about the desired API of `orchard.resource`, I guess at some point we will need to decide whether `orchard` knows or not about the `ops`. In particular I ported this guy this way:

```clj
{:resources-list
   (map #(select-keys % [:file :relpath])
        (get-project-resources))}
```

cause this is what `cider-nrepl` returns - a map containing the resources. Should `orchard` return only the resources or does it know that it needs wrapping?

This problem did not really manifest itself in `info` and I want to give it more thought (and ask for feedback). For now `orchard` will return things "unwrapped".

arichiardi:
That originated my split I would say, I wanted to have one namespace with only the `op` API, as you noted @bbatsov this might not be what we want.

bbatsov:
> That originated my split orchard.resource/orchard.java.resource. I wanted to have one namespace with only the op API, as you noted @bbatsov this might not be what we want.

Orchard should know 0 of `cider-nrepl` and the ops there. If some op needs to process further the data from Orchard that's up the op.

src/orchard/resource.clj
```diff
@@ -0,0 +1,73 @@
...
+          (remove #(re-matches #".*\.(clj[cs]?|java|class)" (:relpath %)))))
+   (filter (memfn isDirectory) (map io/as-file (cp/classpath (cp/boot-aware-classloader))))))
+
+(defn resource-full-path [relative-path]
+  (io/resource relative-path (cp/boot-aware-classloader)))
+
+(defn resource-path-tuple
+  "If it's a resource, return a tuple of the relative path and the full
+  resource path."
+  [x]
```
bbatsov: I guess we can do better than `x` here. :-)

src/orchard/resource.clj
```diff
@@ -0,0 +1,73 @@
...
+      (if-let [[_ relative] (re-find #".*jar:(.*)" x)]
+        (if-let [full (resource-full-path relative)]
+          [relative full]))))
+
+(defn resource-path
+  "Return the resource path for the given input map.
+
+  It requires the :name key."
+  {:added "0.5.0"}
+  [{:keys [name]}]
```
bbatsov: Why does this take a map?

src/orchard/resource.clj
```diff
@@ -0,0 +1,73 @@
...
+  (some-> name (resource-full-path) (.getPath)))
+
+(defn resource-maps
```
bbatsov: I guess that's something needed just for `cider-nrepl`, right? If so - it should go live there.

src/orchard/resource.clj
```diff
@@ -0,0 +1,73 @@
...
+   [orchard.classpath :as cp]))
+
+(defn- trim-leading-separator
+  "Trim the java.io.File/separator at the beginning of s."
+  [s]
+  (if (.startsWith s java.io.File/separator)
+    (subs s 1)
+    s))
+
+(defn- project-resources
```
bbatsov: Probably this should be public. Seems like an unless information.

bbatsov:
Btw, I've noticed some public functions don't have unit tests.

src/orchard/resource.clj
arichiardi: this map was originally the message, but given the new take on things I will just use the name

src/orchard/resource.clj
arichiardi: this is a bigger conversation and I will open an issue, but if we do that, `orchard` won't be the self-contained series of APIs that I think we should strive for. Please bear with me :smile:

arichiardi:
> Orchard should know 0 of cider-nrepl and the ops there. If some op needs to process further the data from Orchard that's up the op.

Fundamentally agree with this, but also opened an issue witn some thoughts in `cider-nrepl`.

arichiardi:
Comments addressed

bbatsov:
Thanks! I'll cut a new beta today, so you can integrate this with `cider-nrepl`.

PR #59, opened 2019-06-23
Correct namespace prefix for inlined deps

dpsutton:

- [ ] All tests are passing

Apropos is broken right now and including inlined deps for this reason and another reason that will be patched on cider-nrepl and cider soon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### Bugs fixed
 
+* [#59](https://github.com/clojure-emacs/orchard/pull/59): Correct inlined-dep prefix for removing namespaces
 * [#57](https://github.com/clojure-emacs/orchard/pull/57): Fix parsing of Java versions for Java 11+.
 
 ## 0.4.0 (2019-01-14)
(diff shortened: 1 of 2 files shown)


src/orchard/xref.clj
```diff
@@ -47,7 +47,9 @@
   {:added "0.5"}
   [var]
   (let [var (as-var var)
-        all-vars (q/vars {:ns-query {:project? true} :private? true})
+        all-vars (q/vars {:ns-query {:project? true
+                                     :exclude-regexps ["^cider.nrepl.inlined-deps" "^nrepl"]}
```
bbatsov: I think we should probably put the list of exclusions in some var that we can just check where this is needed. 

I also wonder if it won't better for the query methods to just filter those out automatically.

dpsutton: The `q/namespace` function does filter it out. It just had the wrong prefix. This technically isn't needed.

dpsutton: I reset the branch to just the single commit and force pushed. no longer there.

bbatsov:
Thanks!

PR #63, opened 2019-07-05
[Fix #61] Avoid processing file info if there is no file entry

### LINKED ISSUE #61: Avoid processing file info if there is no file entry
Just reporting here what has been discovered in https://github.com/clojure-emacs/cider-nrepl/issues/621 for traceability.

From @xificurC's comment:

> I see I forgot that `info*` calls `clj-meta` which calls into various `orchard.meta` functions depending on the supplied map. I recall I had crashes yesterday when my cursor was on `finally`:
> 
> ```clojure
> cider.nrepl.inlined-deps.orchard.v0v5v0-beta8.orchard.info> (clj-meta {:dialect :clj :ns 'clojure.core :sym 'finally})
> {:forms [(try expr* catch-clause* finally-clause?)],
>  :doc
>  "catch-clause => (catch classname name expr*)\n  finally-clause => (finally expr*)\n\n  Catches and handles Java exceptions.",
>  :name finally,
>  :special-form true,
>  :url "https://clojure.org/special_forms#finally"}
> ```
> 
> There's no `:file` there. Based on this observation I would say the `(update :file` code is incorrect and should be checked whether `:file` is actually not nil.

Thanks to him now we have an easy fix, plus I will add a test :smile:

arichiardi:
There was a `NullPointerException` caused by an unnecessary call to
`cp/classpath-file-relative-path`. That code path is now avoided and as a bonus
the file key is not merged at all.

- [X] All tests are passing

### DIFF
--- a/test/orchard/info_test.clj
+++ b/test/orchard/info_test.clj
@@ -41,8 +41,7 @@
                :package orchard.test_ns
                :super java.lang.Object
                :interfaces (clojure.lang.IType)
-               :javadoc "orchard/test_ns/TestType.html"
-               :file nil}
+               :javadoc "orchard/test_ns/TestType.html"}
              (select-keys i [:ns :name :class :package :super :interfaces :arglists :javadoc :file]))))))
 
 (deftest info-defrecord-test
@@ -72,8 +71,7 @@
                               clojure.lang.IPersistentMap
                               java.util.Map
                               java.io.Serializable)
-                 :javadoc "orchard/test_ns/TestRecord.html"
-                 :file nil}
+                 :javadoc "orchard/test_ns/TestRecord.html"}
                (select-keys i [:ns :name :class :package :super :interfaces :arglists :javadoc :file])))))))
 
 (deftest info-special-form-test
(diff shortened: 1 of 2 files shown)


arichiardi:
@bbatsov pong :smiley: 

Worked on this little fix, will get back to the other one soon-ish :smile:

test/orchard/info_test.clj
```diff
@@ -368,6 +366,25 @@
...
+                     :name finally,
+                     :special-form true,
+                     :url "https://clojure.org/special_forms#finally"}]
+      (testing "- boot project"
+        (with-redefs [orchard.misc/boot-project? (constantly true)]
+          (let [i (info/info* params)]
+            (is (= expected (select-keys i [:ns :name :doc :forms :special-form :url])))
+            (is (nil? (:file i))))))
+
+      (testing "- no boot project"
```
bbatsov: Is it just me or are there no differences in the two tests expect the type of the project?

arichiardi: uhm, so I am testing that the `:file` key is not present AND that of course no exception is thrown.

There is a `with-redefs` for `boot` projects and I tested with and without the patch (written the tests first).

Do you have any suggestion for improving it or maybe I am missing something?

bbatsov: I just assumed that because you factored in the project type you were expecting it to be meaningful to those tests. That's why I asked.

arichiardi: Yes the test is making sure we don't have exceptions in both with and without boot

bbatsov: OK, fair enough.

PR #64, opened 2019-07-10
Add finding docs functionality from ClojureDocs

liquidz:
PR for https://github.com/clojure-emacs/cider/issues/2663
Introduce finding docs functionality from not Grimoire but ClojureDocs

- [x] All tests are passing


src/orchard/clojuredocs.clj
```diff
@@ -0,0 +1,35 @@
+(ns orchard.clojuredocs
+  "Find docs from ClojureDocs and retrieve the result as a map."
+  {:author "Masashi Iizuka"}
```
bbatsov: Please, add to the metadata that the namespace was added in version 0.5.

src/orchard/clojuredocs.clj
```diff
@@ -0,0 +1,35 @@
...
+  (:require
+   [clojure.edn :as edn]))
+
+(def cache (atom {}))
+
+(defn- map-from-key [f coll]
+  (reduce #(assoc %1 (f %2) %2) {} coll))
+
+(defn update-documents-cache!
```
bbatsov: I guess we can just name this `update-cache!`, as it's clear from the namespace which cache we are referring to.

src/orchard/clojuredocs.clj
```diff
@@ -0,0 +1,35 @@
...
+  [export-edn-file]
+  (let [docs (-> export-edn-file slurp edn/read-string)]
+    (->> (:vars docs)
+         (group-by #(:ns %))
+         (reduce-kv #(assoc %1 %2 (map-from-key :name %3)) {})
+         (hash-map export-edn-file)
+         (reset! cache))
+    true))
+
+(defn find-document
```
bbatsov: I wonder if we can't come up with a better name here, as the word `document` implies something else. Perhaps `find-doc` (for `documentation`). Can't think of a great name, but names are important.

bbatsov:
Apart from my small remarks the changes look good to me. I just wonder if now we should replace the usage of the `see-also.edn` with this and weather to integrate more of this data (e.g. notes and examples) into the `info` as well.  //cc @arichiardi @SevereOverfl0w @dpsutton

SevereOverfl0w:
It would be nice to carry this along with the info as well, given this is cached it's probably going to have a reasonable performance. We are loading the full 2.8M string into memory with the `slurp`, so that might be something to consider changing, but I haven't looked deeply at the memory meter and considered whether we care.

liquidz:
@bbatsov Thank you for your reviewing!
I fixed them.

arichiardi:
The patch looks good to me apart from the slurp of the whole thing which we could maybe do lazily ad hoc if we had a different data format (one file per ClojureDocs entry?). Could be an optimization for later.

On another note we should potentially merge this to  https://github.com/cljs/api for ClojureScript support.

src/orchard/clojuredocs.clj
```diff
@@ -0,0 +1,36 @@
...
+         (reset! cache))
+    true))
+
+(defn find-doc
+  "Find a document matching to ns-name and var-name from cached documents.
+  If there are no cached documents corresponding to `export-edn-file`, cache will be updated.
+
+  Return nil if there is no matching document."
+  {:added "0.5.0"}
+  [export-edn-file ns-name var-name]
```
bbatsov: One final remark - I think here `edn-file` should be an optional param, as we know the default location of the file after it's downloaded and specifying it explicitly all the time seems redundant to me.

src/orchard/clojuredocs.clj
```diff
@@ -0,0 +1,36 @@
...
+
+(defn- map-from-key [f coll]
+  (reduce #(assoc %1 (f %2) %2) {} coll))
+
+(defn update-cache!
+  "Load exported documents file from ClojureDocs, and store it as a cache.
+  A EDN format file is expected to the `export-edn-file` argument.
+  (e.g. https://clojuredocs-edn.netlify.com/export.edn)"
+  {:added "0.5.0"}
+  [export-edn-file]
```
bbatsov: I think we should make this param optional and load the cache straight from the default url.

bbatsov:
@liquidz After reading the code once more I thought of something else - it'd be nice if we were saving the export file locally, so it could be loaded directly on future invocations of Orchard. We can add some `load-cache` function which checks if the file is locally available (and it's not older than a week) and this function can call `update-cache!` then. I think that'd be a more user-friendly experience in general.

liquidz:
@bbatsov I thought to do that on the editor plugin side.
But because it adds extra work, I agree to save the exported file locally on orchard side.

How about to save the file in `$XDG_CACHE_HOME/orchard/clojuredocs/export.edn` ?
cf. https://standards.freedesktop.org/basedir-spec/latest/

bbatsov:
Yeah, that sounds good to me.

SevereOverfl0w:
https://github.com/soc/directories-jvm even if not used, should have it's methods ported, in order to properly support windows & mac properly.

liquidz:
@SevereOverfl0w Thanks!
I'll port some of them!

liquidz:
@bbatsov @SevereOverfl0w Fixed!

I tested `orchard.directory/cache-dir` for Windows on AppVeyor.
https://ci.appveyor.com/project/liquidz/windows-test/builds/26108911
Tested codes (same as `orchard.directory/cache-dir`) are here:
https://github.com/liquidz/windows-test/blob/master/src/windows_test/directory.clj

src/orchard/directory.clj
```diff
@@ -0,0 +1,67 @@
+(ns orchard.directory
```
bbatsov: I think this namespace should be named in some more generic way (provided we want to keep there os-related functionality). Maybe `orchard.os` or `orchard.platform`? Not sure how this is typically named.

src/orchard/misc.clj
```diff
@@ -3,8 +3,19 @@
    [clojure.java.io :as io]
    [clojure.string :as str]))
 
+(def os-name
```
bbatsov: I think we can put the os-* in the new namespace as well. We'll just keep the existing functions here for compatibility and deprecate them.

src/orchard/directory.clj
```diff
@@ -0,0 +1,67 @@
...
+  "A port of BaseDirectories.java in soc/directories-jvm.
+  https://github.com/soc/directories-jvm"
+  {:author "Masashi Iizuka"
+   :added "0.5.0"}
+  (:require [clojure.java.io :as io]
+            [clojure.string :as str]
+            [orchard.misc :as u])
```
bbatsov: I think `misc` is a better alias than `u`. :-) The weird aliases in the codebase today are due to the fact that this namespace was named `util` in cider-nrepl (if I recall correctly that is).

src/orchard/directory.clj
liquidz: I like `orchard.os` shorter!

liquidz:
@bbatsov Thank you for your reviewing! I fixed them!

bbatsov:
Great work! I'll cut a new beta now, so you can update `cider-nrepl` accordingly.

bbatsov:
Hmm, seems I pushed beta-9 from my other PC without doing `git push` and now I don't have access to the other computer. Guess we'll mess up a bit the git history now, but whatever.

bbatsov:
0.5.0-beta10 is out!

liquidz:
Thanks! I'll create a PR to cider-nrepl.

PR #69, opened 2019-09-01
Include specs in var-meta

dpsutton:
Wasn't threaded before so the value was computed and thrown away.

- [x] All tests are passing

Thanks!

### DIFF
--- a/test/orchard/meta_test.clj
+++ b/test/orchard/meta_test.clj
@@ -78,6 +78,11 @@
 (deftest var-meta-test
   ;; Test files can't be found on the class path.
   (is (:file (m/var-meta #'m/var-meta)))
+  (testing "Includes spec information"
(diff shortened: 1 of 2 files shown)


test/orchard/meta_test.clj
```diff
@@ -2,7 +2,9 @@
   (:require
    [orchard.meta :as m]
    [clojure.repl :as repl]
-   [clojure.test :refer :all]))
+   [clojure.test :refer :all]
+   [clojure.set :as set]
+   [clojure.spec.alpha :as s]))
```
bbatsov: The spec code should be conditionally added, as we still support Clojure 1.8.

dpsutton: yeah just saw the test matrix failure. I'll get to this tonight.

dpsutton:
@bbatsov good to go

bbatsov:
Thanks!

Issue #71, opened 2019-09-17
Handle lack of internet connectivity when the ClojureDocs cache is missing


bbatsov:
A couple of people have experienced problems trying to use `cider-nrepl` when not connected to the internet or behind a corporate firewall, because the download of the ClojureDocs export locks up. We should figure out some solution to this. Some options:

* Do the download in the background
* Try to detect if we have access to the remote host
* Ship some fallback data with orchard

@liquidz Any thoughts on this?

liquidz:
@bbatsov 

> Try to detect if we have access to the remote host

I think this option is good.

> Do the download in the background

With this option, there is a status management during downloading, so codes seems to be complicated.

> Ship some fallback data with orchard

It looks good, but it seems to be painful to update fallback data regularly.
It may be possible to adopt it once and consider the automatic update method later.
BTW I think detecting remote host access will be necessary to fallback after all.

bbatsov:
I don't have a strong preference for how to solve this, I just want us to come up with something that avoids the lock ups.

liquidz:
@bbatsov
OK. I'll try to fix it this weekend!

bbatsov:
Thanks!

SevereOverfl0w:
We should perhaps think about tethered internet case also:

1. Slow
2. Expensive

bbatsov:
Is there even a way to detect this programatically?

bbatsov:
Fixed in #72.

PR #72, opened 2019-09-23
Update to check internet connectivily for clojuredocs

liquidz:

- [x] All tests are passing

Changes for #71

### DIFF
--- a/src/orchard/clojuredocs.clj
+++ b/src/orchard/clojuredocs.clj
@@ -8,7 +8,10 @@
    [clojure.string :as str]
    [orchard.util.os :as os])
   (:import
-   (java.time Instant)))
+   (java.io IOException)
+   (java.net URL)
+   (java.time Instant)
+   (javax.net.ssl HttpsURLConnection)))
 
 (def cache (atom {}))
 (def default-edn-file-url
@@ -18,8 +21,13 @@
                                "orchard"
                                "clojuredocs"
                                "export.edn"]))
-(def ^{:doc "One week. Unit is millisecond."}
-  cache-updating-threshold 604800000)
+(def cache-updating-threshold
+  "One week. Unit is millisecond."
+  604800000)
+
+(def connect-timeout
+  "Timeout value for checking connection. Unit is millisecond."
+  1000)
 
 (defn- write-cache-file! [url]
   (.. (io/file cache-file-name)
@@ -34,6 +42,20 @@
        (reset! cache))
   true)
 
+(defn test-remote-url [^String url]
+  (if-not (.startsWith url "http")
+    [false (IllegalArgumentException. "URL must start wih 'http'.")]
+    (let [url (URL. url)
+          conn ^HttpsURLConnection (.openConnection url)]
+      (.setConnectTimeout conn connect-timeout)
+      (try
+        (.connect conn)
+        [true]
+        (catch IOException ex
+          [false ex])
+        (finally
+          (.disconnect conn))))))
+
 (defn load-cache!
   "Load exported docs file from ClojureDocs, and store it as a cache.
   A EDN format file is expected to the `export-edn-url` argument.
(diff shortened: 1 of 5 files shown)


src/orchard/clojuredocs.clj
```diff
@@ -103,7 +145,7 @@
   ([ns sym export-edn-url]
    (if (special-symbol? sym)
      (find-doc "clojure.core" (str sym) export-edn-url)
-     (-> (ns-resolve ns sym) var-name get-doc))))
+     (some-> (try-ns-resolve ns sym) var-name get-doc))))
```
bbatsov: Why kind of exception are you anticipating here?

src/orchard/clojuredocs.clj
```diff
@@ -21,6 +24,9 @@
 (def ^{:doc "One week. Unit is millisecond."}
   cache-updating-threshold 604800000)
 
+(def ^{:doc "Timeout value for checking connection. Unit is millisecond."}
```
bbatsov: Why don't we use inline docstrings?

src/orchard/clojuredocs.clj
```diff
@@ -34,6 +40,18 @@
        (reset! cache))
   true)
 
+(defn test-to-access-url [^String url]
```
bbatsov: I guess this can be `test-remote-url`. Btw, are we certain this is not going to blow up if someone passes a local file URL instead?

bbatsov:
Overall the PR looks good. I left a couple of small inline remarks.

src/orchard/clojuredocs.clj
liquidz: This is the fix for https://github.com/clojure-emacs/cider-nrepl/blob/v0.22.3/src/cider/nrepl/middleware/clojuredocs.clj#L18

src/orchard/clojuredocs.clj
liquidz: You're right. I fixed to validate the format of `url`.

liquidz:
@bbatsov Thank you for your reviewing! I fixed.
I also added some fix for `orchard.meta` because I missed that it is using `orchard.clojuredocs`.

bbatsov:
@liquidz I notice some weird failure in `cider-nrepl` now, but I can't imagine how those changes could have caused it - see https://circleci.com/gh/clojure-emacs/cider-nrepl/1231?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification Any ideas?

liquidz:
@bbatsov Sorry, it seems `orchard.clojuredocs/test-remote-url` is the cause.
https://github.com/clojure-emacs/orchard/blob/v0.5.2/src/orchard/clojuredocs.clj#L46-L47

`cider.nrepl.middleware.clojuredocs-test` uses local file path as a test url, so `orchard.clojuredocs/test-remote-url` returns `false`.
https://github.com/clojure-emacs/cider-nrepl/blob/master/test/clj/cider/nrepl/middleware/clojuredocs_test.clj#L7

There are two way to fix.
1. Fix `orchard.clojuredocs/load-cache!` to check local file readability also.
2. Fix `cider.nrepl.middleware.clojuredocs-test` to skip `orchard.clojuredocs/test-remote-url` like the following.
   * https://github.com/clojure-emacs/orchard/blob/v0.5.2/test/orchard/clojuredocs_test.clj#L42

I think the latter is good because the former loses the meaning of the cache file.

liquidz:
> Fix cider.nrepl.middleware.clojuredocs-test to skip orchard.clojuredocs/test-remote-url like the following.

Oh, this way does not work well because orcahrd is inlined in cider-nrepl...

bbatsov:
> Fix orchard.clojuredocs/load-cache! to check local file readability also.

Or it can skip checks for local files. Not sure what's the best approach here. I guess I might still cut a new `cider-nrepl` despite this failure, as it doesn't see that important, but we'll certainly have to address the problem.

liquidz:
@bbatsov I don't prefer to add judgement conditions only to pass test, so skipping checks for local files looks good to me.

PR #76, opened 2019-11-02
[WIP] Fix bug with no `:file` for `info` on namespaces w/o defs

PEZ:
Sending this PR in to get feedback.

The problem with `:file` entries being `nil` for namespaces where there are no `def`s is that `orchard.meta/ns-meta` relies on `clojure.core/ns-publics` to first pick out some of the `def`:ed vars in order to use `var-meta` on them.

In this PR I use a backup method for when `ns-publics` return an empty map.

However, in `orchard.info-test/file-resolution-no-defs-issue-75-test` I still get a `nil` back for the `:cljs` test. I have yet failed to figure that out. Any feedback on that is very welcome.

I'd also want some general feedback, as I am planning to try contribute to this library more, and I need to learn the best ways to do that. Cheers!

- [x] All tests are passing

### DIFF
--- a/src/orchard/meta.clj
+++ b/src/orchard/meta.clj
@@ -268,11 +268,15 @@
      (meta ns)
      {:ns (ns-name ns)
       :name (ns-name ns)
-      :file (-> (ns-publics ns)
-                first
-                second
-                var-meta
-                :file)
+      :file (or 
+             (-> (ns-publics ns)
+                 first
+                 second
+                 var-meta
+                 :file)
+             (->
+              (ns/canonical-source ns)
+              .getPath))
       :line 1})))
 
 ;;; ## Manipulation
--- a/src/orchard/meta.clj
+++ b/src/orchard/meta.clj
@@ -261,6 +261,17 @@
                    :form form
                    :code code)))))))
 
+(defn ns-file
+  "Finds the path to the file defining this `ns`"
+  [ns]
+  (or (some-> (ns-publics ns)
+              first
+              second
+              var-meta
+              :file)
+      (some-> (ns/canonical-source ns)
+              .getPath)))
+
 (defn ns-meta
   [ns]
   (when ns
@@ -268,15 +279,7 @@
      (meta ns)
      {:ns (ns-name ns)
       :name (ns-name ns)
-      :file (or 
-             (-> (ns-publics ns)
-                 first
-                 second
-                 var-meta
-                 :file)
-             (->
-              (ns/canonical-source ns)
-              .getPath))
+      :file (ns-file ns)
       :line 1})))
 
 ;;; ## Manipulation
--- a/test/orchard/meta_test.clj
+++ b/test/orchard/meta_test.clj
@@ -110,3 +110,14 @@
     (docs/clean-cache!)
     (testing "Including see-also is skipped"
       (is (not (contains? (m/var-meta (resolve 'clojure.set/union)) :see-also))))))
+
+(deftest ns-meta-test
+  (testing "Includes a non-nil :file"
+    (is (some-> 'orchard.test-ns-dep
+                (find-ns)
+                (m/ns-meta)
+                :file))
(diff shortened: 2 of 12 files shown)


src/orchard/meta.clj
```diff
@@ -268,11 +268,15 @@
      (meta ns)
      {:ns (ns-name ns)
       :name (ns-name ns)
-      :file (-> (ns-publics ns)
-                first
-                second
-                var-meta
-                :file)
+      :file (or
```
bbatsov: I guess you can extract this as a function - `ns-file`.

test/orchard/meta_test.clj
```diff
@@ -110,3 +110,14 @@
     (docs/clean-cache!)
     (testing "Including see-also is skipped"
       (is (not (contains? (m/var-meta (resolve 'clojure.set/union)) :see-also))))))
+
+(deftest ns-meta-test
+  (testing "Includes a non-nil :file"
+    (is (some-> 'orchard.test-ns-dep
```
bbatsov: Why the need for `some->` here? After all you expect this not to be `nil`.

bbatsov:
Your changes seem good me to me. Not sure what's causing the failure. Maybe @arichiardi would know.

test/orchard/meta_test.clj
PEZ: It's the whole thread that should return something not `nil`, right? Maybe I do not follow the question...

bbatsov: My bad. Too much multitasking. :D

src/orchard/meta.clj
```diff
@@ -268,11 +268,15 @@
...
-                var-meta
-                :file)
+      :file (or 
+             (-> (ns-publics ns)
+                 first
+                 second
+                 var-meta
+                 :file)
+             (->
+              (ns/canonical-source ns)
```
bbatsov: I guessing when you try this manually it actually works (instead of returning nil), right?

PEZ: Yes, `(ns-meta 'orchard.test-no-defs)` returns the right thing. And the `:clj` ”branch” of the test also passes. But the `:cljs` part of the test does not.

bbatsov: Seems `canonical-source` doesn't work with `cljs` files:

```
(defn canonical-source
  "Returns the URL of the source file for the namespace object or symbol,
  according to the canonical naming convention, if present on the classpath."
  [ns]
  (let [path (-> (str ns)
                 (str/replace "-" "_")
                 (str/replace "." "/"))]
    (or (io/resource (str path ".clj"))
        (io/resource (str path ".cljc")))))
```

PEZ: Good catch! This would have made me fail fix https://github.com/BetterThanTomorrow/calva/issues/427 even when we find out why the cljs tests fail. 😄

PEZ:
Yes, the only failing test. I think I have found another clue now. `ns-meta` is not used for the `:cljs` dialect. 😄 

I am now tracking down what is used there instead. Good to know that the rests looks good.

arichiardi:
Yes the Cljs is more complicated because you need to get to the Compiler state. The wrappers though are are in the `cljs` specific namespace (meta.cljs and analysis.cljs iirc).

Not a the keyboard now but I can help you with that at some point.

arichiardi:
The `clj-meta` and `cljs-meta` are the entry point and we probably should check the fix for both branches of info

PEZ:
Thanks! Found it. Fixed it in a similar way as I did for the `:clj` dialect.

Also, the `info` ns had its own ”copy” of the `normalize-ns-meta` fn. It had evolved past the one in the `cljs.meta` namespace. It took me a while to realise the latter one was not being used. I choose to move the used version to the `cljs.meta` namespace and delete the one in there. So now there is only one. And it has the fix for this no-defs business.

PR #80, opened 2019-12-24
[inspector] Render nil as nil inside data structures

alexander-yakushev:
Before this change, `nil` inside data structures did not render at all. The change does not require changes in cider-nrepl or cider (however, maybe tests would have to be updated when cider-nrepl bumps orchard version).

Before:

image

image

After:

image

image


bbatsov:
Nice!

PR #81, opened 2019-12-24
[inspector] Handle InaccessibleObjectException for foreign-module fields

alexander-yakushev:
On Java9+, when trying to inspect an object of a class that comes from the other module, and that object has private fields, JVM will throw InaccessibleObjectException when the inspector tries to set those private fields to accessible. Unfortunately, there doesn't seem a generic way to work around this, so at least we want the inspector to not crash in such situations.

Here's what the inspector shows after the patch:

image

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -425,11 +425,19 @@
               (.getName f))
 
             (field-val [^Field f]
-              (try (.setAccessible f true)
-                   (catch java.lang.SecurityException e))
-              (try (.get f obj)
-                   (catch java.lang.IllegalAccessException e
-                     "Access denied.")))
+              (let [e (try (.setAccessible f true)
+                           nil
+                           (catch Exception e
+                             ;; We want to handle specifically SecurityException
+                             ;; and j.l.r.InaccessibleObjectException, but the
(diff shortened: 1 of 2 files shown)


src/orchard/inspect.clj
```diff
@@ -425,11 +425,19 @@
...
-                   (catch java.lang.SecurityException e))
-              (try (.get f obj)
-                   (catch java.lang.IllegalAccessException e
-                     "Access denied.")))
+              (let [e (try (.setAccessible f true)
+                           nil
+                           (catch Exception e
+                             ;; We want to handle specifically SecurityException
+                             ;; and j.l.r.InaccessibleObjectException, but the
+                             ;; letter only comes with Java9+, so let's just
```
bbatsov: letter -> latter

bbatsov:
Btw, please mention this in the changelog as well.

src/orchard/inspect.clj
alexander-yakushev: 🤦‍♂

alexander-yakushev:
So, is it good enough now?

bbatsov:
Yep. Thanks!

PR #82, opened 2020-01-31
Add handling for java version strings with $OPT chunks

tstirrat15:
## Motivation
See comment here: https://github.com/clojure-emacs/orchard/pull/57#issuecomment-580518531

It's currently causing cider to fail for me when I jack in, and the traceback includes orchard failing to get the correct version string.

## Changes
* Add a test for a java version string with a $OPT chunk
* Fix the test by lopping off the $OPT in the `let` clause of `parse-java-version`

## Notes
I'm still pretty new to clojure, so if there's a more sane/elegant/appropriate way to lop off the $OPT bit, please let me know.

### DIFF
--- a/src/orchard/misc.clj
+++ b/src/orchard/misc.clj
@@ -92,7 +92,8 @@
   "Parse a Java version string according to JEP 223 and return the appropriate version."
   [java-ver]
   (try
-    (let [[major minor _] (str/split java-ver #"\.")
+    (let [[no-opt _] (str/split java-ver #"-")
+          [major minor _] (str/split no-opt #"\.")
           major (Integer/parseInt major)]
       (if (> major 1)
         major
--- a/src/orchard/misc.clj
+++ b/src/orchard/misc.clj
@@ -92,6 +92,8 @@
   "Parse a Java version string according to JEP 223 and return the appropriate version."
   [java-ver]
   (try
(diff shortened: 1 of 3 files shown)


src/orchard/misc.clj
```diff
@@ -92,7 +92,8 @@
   "Parse a Java version string according to JEP 223 and return the appropriate version."
   [java-ver]
   (try
-    (let [[major minor _] (str/split java-ver #"\.")
+    (let [[no-opt _] (str/split java-ver #"-")
```
bbatsov: I'd probably add some comment here explaining the need for this, as it's exactly obvious.

bbatsov:
The change seems fine to me. It should also be reflected in the changelog as a bugfix.

tstirrat15:
I've added those changes. Let me know if you want me to squash commits or anything like that.

tstirrat15:
What's the status?

bbatsov:
My bad. I thought I had already merged this! Thanks for tackling it. Next week I'll cut some new Orchard release, as this week I'm busy around IN/Clojure.

tstirrat15:
Sounds good. Thank you!

Where should I look for a notification of that new release?

bbatsov:
I think that if you're watching a repo you get release notifications. I don't usually do public release announcements for bugfix releases.

PR #83, opened 2020-02-21
Ignore non file URLs when checking for directory or file extensions

fp7:
In a URLClassLoader not all URLs have to be files or directories.
Eg. Spring boot's fat jars add entries like "jar:file:..." to the
classpath

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* [#83](https://github.com/clojure-emacs/orchard/pull/83): Ignore non file URLs when checking for directory or file extensions.
+
 ## 0.5.6 (2020-02-14)
 
 ### Bugs fixed
--- a/src/orchard/misc.clj
+++ b/src/orchard/misc.clj
@@ -20,16 +20,27 @@
   []
   (not (nil? (boot-fake-classpath))))
 
+(defn url?
+  "Check whether the argument is an url"
+  [u]
+  (instance? java.net.URL u))
+
 (defn directory?
-  "Whether the argument is a directory"
+  "Whether the argument is a directory or an url that points to a directory"
   [f]
-  (.isDirectory (io/as-file f)))
+  (if (url? f)
+    (and (= (.getProtocol ^java.net.URL f) "file")
+         (.isDirectory (io/as-file f)))
+    (.isDirectory (io/as-file f))))
 
 (defn file-ext?
   "Whether the argument's path ends in one of the specified case-insensitive
   file extensions"
   [f & exts]
-  (let [file (io/as-file f)]
+  (when-let [file (if (url? f)
+                    (when  (= (.getProtocol ^java.net.URL f) "file")
+                      (io/as-file f))
+                    (io/as-file f))]
     (some (fn [ext]
             (.endsWith (.. file getName toLowerCase) ext))
           exts)))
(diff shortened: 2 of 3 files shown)


bbatsov:
The proposed change seems reasonable to me. Probably we can make the `if`s a bit more compact, but that's not a big deal. Please, mention your fix in the changelog.

fp7:
How would you make it more compact?

src/orchard/misc.clj
```diff
@@ -20,16 +20,27 @@
...
+         (.isDirectory (io/as-file f)))
+    (.isDirectory (io/as-file f))))
 
 (defn file-ext?
   "Whether the argument's path ends in one of the specified case-insensitive
   file extensions"
   [f & exts]
-  (let [file (io/as-file f)]
+  (when-let [file (if (url? f)
+                    (when  (= (.getProtocol ^java.net.URL f) "file")
```
bbatsov: Here you can combine this condition with the previous using `and` (for instance). Or you can just replace the condition with one `or`/`and`.

src/orchard/misc.clj
```diff
@@ -20,16 +20,27 @@
...
+  [u]
+  (instance? java.net.URL u))
+
 (defn directory?
-  "Whether the argument is a directory"
+  "Whether the argument is a directory or an url that points to a directory"
   [f]
-  (.isDirectory (io/as-file f)))
+  (if (url? f)
+    (and (= (.getProtocol ^java.net.URL f) "file")
```
bbatsov: Here you can pull this `and` higher.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#83](https://github.com/clojure-emacs/orchard/pull/83): Ignore non file URLs when checking for directory or file extensions
```
bbatsov: You should also put "Bugs fixed" above this entry.

bbatsov: And a `.` at the end of this sentence. :-)

src/orchard/misc.clj
fp7: You mean like this ?
```
(and (url? f)
        (= (.getProtocol ^java.net.URL f) "file")
        (.isDirectory (io/as-file f)))
```
Then it would not check nonURLs correctly?

CHANGELOG.md
fp7: Oh yeah, is see. Sorry

src/orchard/misc.clj
bbatsov: I was thinking of something like:

`(and (url? f) (= (.getProtocol ^java.net.URL f) "file")) ...`

With a more complex conditional you can even get rid of the duplicated .isDirectory check, but this will probably make the code hard to read.

fp7: Sorry, but i still do not understand.
You mean like this?
```
(defn directory?
  "Whether the argument is a directory or an url that points to a directory"
  [f]
  (if (and (url? f) (= (.getProtocol ^java.net.URL f) "file"))
    (.isDirectory (io/as-file f))))
```

bbatsov: My bad, I just realized where your confusion was coming from. If we structure the condition in this manner, the else branch won't be correct. I should stop reviewing code while doing something else. :D

bbatsov:
Can you also re-run the failed build? Seems like something unrelated, but still - let's see what happens if we run it a couple of times.

fp7:
Is there a button somewhere that i can click to rerun the tests?

src/orchard/misc.clj
fp7: No problem

src/orchard/misc.clj
fp7: Still something to do here?

PR #84, opened 2020-02-23
Fix javadoc urls on java 11

dpsutton:
the core-java-api url prefix used from clojure.java.javadoc includes
the java.base module so we just strip that out and use the module as
we had been doing

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -363,7 +363,9 @@
       (some (let [classname (.replaceAll path "/" ".")]
               (fn [[prefix url]]
                 (when (.startsWith classname prefix)
-                  (str url path))))
+                  (str (cond-> url
+                         (= 11 misc/java-api-version) (.replaceFirst "/java.base" ""))
+                       path))))
             @javadoc/*remote-javadocs*)
       path))
 
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -373,7 +373,8 @@
                   ;; 8. clojure 1.10.1 doesn't have 13. We just backport them
                   ;; regardless of clojure version
                   (zipmap ["java." "javax." "org.ietf.jgss." "org.omg." "org.w3c.dom." "org.xml.sax"]
-                          (repeat (backported-javadoc-bases misc/java-api-version)))))
+                          (repeat (or (backported-javadoc-bases misc/java-api-version)
+                                      (backported-javadoc-bases 11))))))
       path))
 
 ;;; ## Initialization
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -345,7 +345,7 @@
           (first ms)
           {:candidates (zipmap (map :class ms) ms)})))))
 
-(def backported-javadoc-bases
+(def javadoc-base-urls
   "Copied from clojure.java.javadoc. These are the base urls for
   javadocs from `clojure.java.javadoc/*core-java-api*`. It is here for
   two reasons:
@@ -373,8 +373,8 @@
                   ;; 8. clojure 1.10.1 doesn't have 13. We just backport them
                   ;; regardless of clojure version
                   (zipmap ["java." "javax." "org.ietf.jgss." "org.omg." "org.w3c.dom." "org.xml.sax"]
-                          (repeat (or (backported-javadoc-bases misc/java-api-version)
-                                      (backported-javadoc-bases 11))))))
+                          (repeat (or (javadoc-base-urls misc/java-api-version)
+                                      (javadoc-base-urls 11))))))
       path))
 
 ;;; ## Initialization
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -345,28 +345,48 @@
           (first ms)
           {:candidates (zipmap (map :class ms) ms)})))))
 
+(def backported-javadoc-bases
+  "Copied from clojure.java.javadoc. These are the base urls for
+  javadocs from `clojure.java.javadoc/*core-java-api*`. It is here for
+  two reasons:
+  1. add java 13 to this list
+  2. backport for older Clojures"
+  {8 "https://docs.oracle.com/javase/8/docs/api/"
+   9 "https://docs.oracle.com/javase/9/docs/api/"
+   10 "https://docs.oracle.com/javase/10/docs/api/"
+   11 "https://docs.oracle.com/en/java/javase/11/docs/api/"
+   12 "https://docs.oracle.com/en/java/javase/12/docs/api/"
+   13 "https://docs.oracle.com/en/java/javase/13/docs/api/"})
+
+(def remote-javadocs
+  (sorted-map
+   "com.google.common." "http://google.github.io/guava/releases/23.0/api/docs/"
+   "java." backported-javadoc-bases
+   "javax." backported-javadoc-bases
+   "org.ietf.jgss." backported-javadoc-bases
+   "org.omg." backported-javadoc-bases
+   "org.w3c.dom." backported-javadoc-bases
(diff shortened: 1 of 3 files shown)


src/orchard/java.clj
```diff
@@ -363,7 +363,9 @@
       (some (let [classname (.replaceAll path "/" ".")]
               (fn [[prefix url]]
                 (when (.startsWith classname prefix)
-                  (str url path))))
+                  (str (cond-> url
+                         (= 11 misc/java-api-version) (.replaceFirst "/java.base" ""))
```
bbatsov: Isn't this a problem on newer JDK releases as well?

src/orchard/java.clj
```diff
@@ -363,7 +363,9 @@
       (some (let [classname (.replaceAll path "/" ".")]
               (fn [[prefix url]]
                 (when (.startsWith classname prefix)
-                  (str url path))))
+                  (str (cond-> url
```
bbatsov: Might be a good idea to add some comments explaining the need for the special handling.

bbatsov:
The changes look good. You should also mention the bug fix in the changelog.

dpsutton:
There are some test failures in clojure 1.8. Working through those

src/orchard/java.clj
```diff
@@ -345,28 +345,48 @@
...
+  javadocs from `clojure.java.javadoc/*core-java-api*`. It is here for
+  two reasons:
+  1. add java 13 to this list
+  2. backport for older Clojures"
+  {8 "https://docs.oracle.com/javase/8/docs/api/"
+   9 "https://docs.oracle.com/javase/9/docs/api/"
+   10 "https://docs.oracle.com/javase/10/docs/api/"
+   11 "https://docs.oracle.com/en/java/javase/11/docs/api/"
+   12 "https://docs.oracle.com/en/java/javase/12/docs/api/"
+   13 "https://docs.oracle.com/en/java/javase/13/docs/api/"})
```
bbatsov: Probably we should also add some fallback for unreleased versions - I assume they are going to have the same URL format as JDK 11+.

src/orchard/java.clj
```diff
@@ -345,28 +345,48 @@
...
-      (when (re-find #"^(java|javax|jdk|org.omg|org.w3c.dom|org.xml.sax)/" path)
-        (apply str ["https://docs.oracle.com"
-                    (if (>= misc/java-api-version 11) "/en/java/javase/" "/javase/")
-                    misc/java-api-version
-                    "/docs/api/"
-                    path]))
-      ;; If that didn't work, _then_ we fallback on `*remote-javadocs*`.
       (some (let [classname (.replaceAll path "/" ".")]
-              (fn [[prefix url]]
+              (fn [[prefix url|version->url]]
```
bbatsov: Haven't seen such a naming convention yet. :-)

src/orchard/java.clj
dpsutton: yeah. the latest version fixes this up.

src/orchard/java.clj
dpsutton: I looked at 14 is quite different: 

14:
https://download.java.net/java/GA/jdk14/docs/api/java.base/java/lang/String.html 

13:
https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/String.html

I'm not sure if 14 is still in some kind of preview or if this is a new schema

bbatsov:
Seems something's broken for Java 8.

dpsutton:
Fixed! Thanks @bbatsov for looking over this. Should be good to go.

src/orchard/java.clj
```diff
@@ -345,26 +345,36 @@
           (first ms)
           {:candidates (zipmap (map :class ms) ms)})))))
 
+(def backported-javadoc-bases
```
bbatsov: Probably we can name this `javadoc-base-urls`. I think `backported` is a bit confusing in this context.

bbatsov:
@dpsutton Looks good to me. Just check my tiny remark, add some changelog entry and squash the related commits together.

bbatsov:
Seems you also have to rebase on top of `master`.

affan-salman:
Hi @dpsutton, is this complete from your perspective?  I'm working on a Java-interop-heavy project and will appreciate early access if this can't be released soon.

affan-salman:
Hi @bbatsov, while I'm not familiar with orchard, I'm happy to test, provide feedback or help out if it can be useful.

dpsutton:
@affan-salman I think its done but being in orchard, it needs a new release here and then a release of cider-nrepl to require it.

In the meantime you can test it and use it locally by the following:

```clj
user> (apropos "resolve-javadoc")
(cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java/resolve-javadoc-path)
user> (dir cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java)
Reflected
cache
class-info
class-info*
javadoc-url
jdk-find
jdk-sources
jdk-tools
member-info
module-name
reflect-info
resolve-class
resolve-javadoc-path
resolve-member
resolve-symbol
source-info
type-info
typesym
nil
user> (in-ns 'cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java)
#namespace[cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java]
cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java> 
(def javadoc-base-urls
  "Copied from clojure.java.javadoc. These are the base urls for
  javadocs from `clojure.java.javadoc/*core-java-api*`. It is here for
  two reasons:
  1. add java 13 to this list
  2. backport for older Clojures"
  {8 "https://docs.oracle.com/javase/8/docs/api/"
   9 "https://docs.oracle.com/javase/9/docs/api/"
   10 "https://docs.oracle.com/javase/10/docs/api/"
   11 "https://docs.oracle.com/en/java/javase/11/docs/api/"
   12 "https://docs.oracle.com/en/java/javase/12/docs/api/"
   13 "https://docs.oracle.com/en/java/javase/13/docs/api/"})

(defn resolve-javadoc-path
  "Resolve a relative javadoc path to a URL and return as a map. Prefer javadoc
  resources on the classpath; then use online javadoc content for core API
  classes. If no source is available, return the relative path as is."
  [^String path]
  (or (resource/resource-full-path path)
      (some (let [classname (.replaceAll path "/" ".")]
              (fn [[prefix url]]
                (when (.startsWith classname prefix)
                  (str url path))))
            (into @javadoc/*remote-javadocs*
                  ;; clojure 1.8 has no javadoc for anything beyond java
                  ;; 8. clojure 1.10.1 doesn't have 13. We just backport them
                  ;; regardless of clojure version
                  (zipmap ["java." "javax." "org.ietf.jgss." "org.omg." "org.w3c.dom." "org.xml.sax"]
                          (repeat (or (javadoc-base-urls misc/java-api-version)
                                      (javadoc-base-urls 11))))))
      path))
#'cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java/javadoc-base-urls#'cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java/resolve-javadoc-path
cider.nrepl.inlined-deps.orchard.v0v5v6.orchard.java> 
```

affan-salman:
Fantastic @dpsutton!  I've tested atop AdoptOpenJDK 11.0.6 and I was able to see javadoc from docs.oracle.com for classes under java.util instead of seeing HTTP 404 without this fix.

 The same test also worked correctly with AdoptOpenJDK 1.8.0_242.

dpsutton:
@bbatsov ping :) should be good to merge.

bbatsov:
Thanks for the reminder! :-)

PR #89, opened 2020-04-21
Fix clj meta java symbol

arichiardi:

- [X] The commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md)
- [X] All tests are passing
- [X] The new code is not generating reflection warnings

Here is the fix that seems to work. Please read the commit message, this is becoming black magic at this point and probably I will need to refactor stuff at some point.
I left the previous function name mostly untouched but given they return `info` candidates I really think that the `java/resolve-` prefix does not make too much sense. They probably belong to `orchard.info` but maybe this deserves another PR.

### DIFF
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -323,27 +323,42 @@
          (filter identity)
          (distinct))))
 
+(defn trim-one-dot
+  [s]
+  (str/replace s #"^\.|\.$" ""))
+
 (defn resolve-symbol
-  "Given a namespace and a class or member symbol, resolve the class/member.
-  Class symbols, constructors, and static calls are resolved to the class
-  unambiguously. Instance members are resolved unambiguously if defined by only
-  one imported class. If multiple imported classes have a member by that name, a
-  map of class names to member info is returned as `:candidates`."
+  "Return the info map for a Java member symbol.
+
+  Constructors and static calls are resolved to the class
+  unambiguously. Instance members are resolved unambiguously if defined
+  by only one imported class. If multiple imported classes have a member
+  by that name, a map of class names to member info is returned as
+  `:candidates`."
   [ns sym]
   {:pre [(every? symbol? [ns sym])]}
-  (let [name (-> (str sym)
-                 (str/replace #"^\.|\.$" "")) ; strip leading/trailing dot
-        sym* (symbol name)
-        [class static-member] (->> (str/split name #"/" 2)
+  (let [sym (-> sym str trim-one-dot)
+        sym* (symbol sym)
+        [class static-member] (->> (str/split sym #"/" 2)
                                    (map #(when % (symbol %))))]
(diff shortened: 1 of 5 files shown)


arichiardi:
@bbatsov I need a lil' review on this

bbatsov:
Yeah, certainly. I took a quick look at this PR a few days ago and then I forgot about it. I recall I was a bit confused about the nature of the fix, but it seems it's mostly a matter of doing some checks in a different order. I'll review this more carefully today or tomorrow. A changelog entry is needed, btw.

src/orchard/java.clj
```diff
@@ -323,28 +323,43 @@
          (filter identity)
          (distinct))))
 
+(defn trim-one-dot
```
bbatsov, line 326: Probably this shouldn't be public API. I'm also not sure why exactly do you need this. Remove `.` from constructors?

src/orchard/java.clj
```diff
@@ -323,28 +323,43 @@
...
+      (when static-member
+        (member-info (:class c) static-member))     ; SomeClass/methodCall
       (when-let [ms (seq (resolve-member ns sym*))] ; methodCall
         (if (= 1 (count ms))
           (first ms)
           {:candidates (zipmap (map :class ms) ms)})))))
 
+(defn resolve-type
+  "Return type info, for a Java class, interface or record."
+  [ns sym]
```
bbatsov, line 354: Probably `sym` is not the best param here.

test/orchard/info_test.clj
```diff
@@ -425,6 +429,19 @@
 (deftest info-java-test
   (is (info/info-java 'clojure.lang.Atom 'swap)))
 
+(deftest info-java-member-precendence-test
+  (testing "Integer/max - issue #86"
```
bbatsov, line 433: I can't understand what this description means. :-)

src/orchard/java.clj
arichiardi: Uhm, ok Will think about a better name

src/orchard/java.clj
arichiardi: Yep, it was already there, I have just factored it out. I will make it private.

arichiardi:
Yes your analysis is correct, we need a different order and interleave. The commit message should contain this info but please let me know if not clear.

bbatsov:
> The commit message should contain this info but please let me know if not clear.

Yeah, it does, but it's also nice to be able to infer all of this by reading the code, docstrings, etc.

arichiardi:
Ooops sorry, got busy, thanks for merging!

bbatsov:
@arichiardi Follow up PRs are always welcome as well. ;-) I just plan to cut a new release later this week, so some patch is better than no patch. :-)

PR #93, opened 2020-06-20
WIP - Retrieve java classes with resolve and fix #86 again

### LINKED ISSUE #86: Faulty logic in info op
## Expected behavior

When calling `orchard.info/info` on a Java member or non-existent fully qualified symbol, the unqualified symbol should not be used to resolve a var in the current namespace.

```clj
(orchard.info/info 'user 'Integer/max)
(orchard.info/info 'user 'non-existing-ns/get)
```
should return the info for the `Integer.max` Java static member and `nil` respectively.

## Actual behavior
Instead the unqualified symbols `max` and `get` are resolved to the clojure.core functions and produces an incorrect info response.

I traced it to this line in clj-meta: https://github.com/clojure-emacs/orchard/blob/ccf90fe120e764aee6b8543d07a103cd6767d85b/src/orchard/info.clj#L76 

Even though `sym` is fully qualified, the processing step `normalise-params` added the key `:unqualified-sym`. I don't know enough to say if it's a predecence issue or if there should be an additional condition to not resolve unqualified-sym if the sym is fully qualified.

In Emacs, call `cider-doc` on the symbol `Integer/max` or observe the eldoc which displays the wrong arglists for `clojure.core/max`.

### Clojure version

E.g. 1.10.2

### Java version

E.g. 1.11

### Operating system

macOS 10.15

arichiardi:

- [ ] All tests are passing

Thanks!


src/orchard/info.clj
```diff
@@ -64,8 +64,11 @@
 
 (defn clj-meta
   {:added "0.5"}
-  [{:keys [dialect ns sym computed-ns unqualified-sym]}]
+  [{:keys [dialect ns sym computed-ns unqualified-sym] :as params}]
   {:pre [(= dialect :clj)]}
+  (println "###")
```
bbatsov, line 69: It's much easier to debug with the debugger. ;-)

bbatsov:
Any updates here?

PR #101, opened 2020-09-30
Remove reflection warnings

BrunoBonacci:
It removed all the remaining reflection warnings.

---

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -394,7 +394,7 @@
 (defmethod inspect :array [inspector obj]
   (-> inspector
       (render-labeled-value "Class" (class obj))
-      (render-labeled-value "Count" (alength obj))
+      (render-labeled-value "Count" (java.lang.reflect.Array/getLength obj)) ; avoid reflection warning from Clojure compiler
       (render-labeled-value "Component Type" (.getComponentType (class obj)))
       (render-ln "Contents: ")
       (render-collection-paged obj)))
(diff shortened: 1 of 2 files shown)


BrunoBonacci:
The failing test is unrelated to the changes and seems to be related to a known issue.

src/orchard/inspect.clj
```diff
@@ -394,7 +394,7 @@
 (defmethod inspect :array [inspector obj]
   (-> inspector
       (render-labeled-value "Class" (class obj))
-      (render-labeled-value "Count" (alength obj))
+      (render-labeled-value "Count" (java.lang.reflect.Array/getLength obj))
```
bbatsov, line 397: I guess this might be confusing to someone without a comment explaining why we don't just use `alength`.

src/orchard/meta.clj
```diff
@@ -234,14 +234,18 @@
               collect? (atom false)
               pbr      (proxy [LineNumberingPushbackReader] [rdr]
                          (read []
-                           (let [i (proxy-super read)]
+                           (let [;; remove reflection warning on proxy-super call
```
bbatsov, line 237: I guess inline comments would be better, as they don't break the code flow here.

BrunoBonacci: I put this on the line above to avoid a long line (120 chars), however, if you prefer them online I'll fix them.

src/orchard/inspect.clj
BrunoBonacci: fair, added a comment.
Again, I would have put this in the line above, but as per your other comment, i've added an inline note.

PR #108, opened 2021-03-15
Add clj-kondo

solar05:

- [x] All tests are passing

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,10 @@ commands:
           command: |
             sudo apt-get install make
       - run:
+          name: Install clj-kondo
+          command: |
+           sudo curl -sLO https://raw.githubusercontent.com/clj-kondo/clj-kondo/master/script/install-clj-kondo && sudo chmod +x install-clj-kondo && sudo ./install-clj-kondo
+      - run:
           name: Generate Cache Checksum
           command: |
             for file in << parameters.files >>
@@ -201,6 +205,10 @@ workflows:
                 name: Running cljfmt
                 command: |
                   make cljfmt
+            - run:
+                name: Running clj-kondo
+                command: |
+                  make kondo
       # - util_job:
       #     name: Code coverage
       #     steps:
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ eastwood:
 cljfmt:
 	lein with-profile +$(VERSION),+cljfmt cljfmt check
 
+kondo:
+	clj-kondo --lint src
+
 # Cloverage can't handle some of the code in this project.  For now we
 # must filter problematic namespaces (`-e`) and tests (`-t`) from
 # instrumentation. Note: this means for now coverage reporting isn't
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ cljfmt:
 	lein with-profile +$(VERSION),+cljfmt cljfmt check
 
 kondo:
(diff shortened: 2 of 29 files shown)


Makefile
```diff
@@ -23,6 +23,9 @@ eastwood:
 cljfmt:
 	lein with-profile +$(VERSION),+cljfmt cljfmt check
 
+kondo:
+	clj-kondo --lint src
```
vemv, line 27: `test` also deserves linting

(same in https://github.com/clojure-emacs/cider-nrepl/)

bbatsov: Agreed.

src/orchard/java/legacy_parser.clj
```diff
@@ -69,7 +68,6 @@
   (when-let [res (io/resource path)]
     (let [access   (ModifierFilter. ModifierFilter/ALL_ACCESS)
           context  (doto (Context.) (Messager/preRegister "orchard-javadoc"))
-          options  (doto (Options/instance context) (.put "ignore.symbol.file" "y"))
```
bbatsov, line 72: It seems to me this might have been invoked for its side effects, not for the result. Notice the `.put`.

bbatsov: On the other hand if those options are not used anywhere - perhaps they are not really needed. @jeffvalk is the expert on this part of the code.

bbatsov:
Let's also run `kondo` on the tests and we're good to good.

test/orchard/apropos_test.clj
```diff
@@ -2,8 +2,8 @@
   (:require
    [clojure.repl :as repl]
    [clojure.string :as str]
-   [clojure.test :refer :all]
-   [orchard.apropos :refer :all]
+   [clojure.test :refer [are deftest is testing]]
+   [orchard.apropos :refer [find-symbols]]
```
vemv, line 6: Perhaps it'd be more idiomatic to:

```diff
-orchard.apropos :refer [find-symbols]]
+orchard.apropos :as sut]
```

Refs:

https://en.wikipedia.org/wiki/System_under_test
https://github.com/clojure-emacs/clj-refactor.el/blob/9f88174878d62e3906be5b04b8ba7788e6ca4570/clj-refactor.el#L220

.circleci/config.yml
```diff
@@ -62,6 +62,10 @@ commands:
           name: Install make
           command: |
             sudo apt-get install make
+      - run:
+          name: Install clj-kondo
+          command: |
+           sudo curl -sLO https://raw.githubusercontent.com/clj-kondo/clj-kondo/master/script/install-clj-kondo && sudo chmod +x install-clj-kondo && sudo ./install-clj-kondo
```
vemv, line 68: This would be a good chance to review whether https://github.com/clojure-emacs/cider-nrepl/pull/693 is a desirable technique

vemv (commented):
Left some suggestions, but don't take my word for them :)

test/orchard/apropos_test.clj
bbatsov: I agree this makes sense in general, but I think it's ortogonal to this PR.

.circleci/config.yml
bbatsov: I think it is, so you're welcome to apply it here as well.

bbatsov:
Thanks!

PR #110, opened 2021-04-07
Address reflection warnings + Upgrade Eastwood

vemv:
> See also: https://github.com/clojure-emacs/cider-nrepl/pull/692#issuecomment-814643076

image


eastwood.clj
```diff
@@ -1,49 +0,0 @@
-(disable-warning
```
vemv, line 1: Proposing to remove this file because the linter passes with zero warnings without it.

Eastwood, because of its approach aims to be an accurate linter that needs as few workarounds as possible.

If experiencing an issue, don't hesitate reporting it upstream and we'd be pleased to either provide better defaults, or tackle false positives as such :)

src/orchard/java/legacy_parser.clj
```diff
@@ -84,7 +84,7 @@
                                         .getKind
                                         .asInterface)
                                     ClassTree))
-                        (into-array)
+                        ^"[Ljava.lang.Object;" (into-array)
```
vemv, line 87: see e.g. https://github.com/JetBrains/jdk8u_langtools/blob/c6b4d9120b96cc4a9c1790236e0abbb79e0e74ce/src/share/classes/com/sun/tools/javac/util/List.java#L161

bbatsov:
Thanks!

Issue #103, opened 2021-01-05
Modifications to classpath by middleware are not visible to user nREPL session


jeffvalk:
### Summary

Middleware is loaded in a different classloader context than user interactions with that middleware. The latter shares no modifiable classloader parent with the former, so changes made to the classpath when middleware is loaded are not seen by user interactions.

### Impact
The primary example of this is in `java.clj`, which adds the JDK `src.zip` to the classpath and sets the `jdk-sources` variable to the zip file's path if found. In-editor JDK class/method docstrings (and argument names, etc) depends on this feature, as does source navigation (jumping) for JDK classes.

The following issues appear to describe this bug:

- https://github.com/clojure-emacs/cider/issues/2732
- https://github.com/clojure-emacs/cider/issues/2687
- https://github.com/clojure-emacs/cider-nrepl/issues/687

### Scope

This behavior happens on both JDK8 and 9+. `dynapath` formerly hacked `java.net.URLClassLoader` to allow it to be modifiable on JDK8 and below, which would have prevented this issue on JDK8 (albeit in an unsupported way), but recent versions of `dynapath` removed this hack, so all JDK versions behave the same wrt this issue.

### Investigation

Below are the values of `jdk-sources`, the classpath, and the classloader hierarchy at middleware load time (via print statements in the `java` middleware), and at runtime in a user REPL session. Note that:

1. Both contexts reflect that `src.zip` has been found (`jdk-sources` is set); however
2. `src.zip` is visible on the classpath at load time, but not at runtime in the REPL session, because 
3. there is no shared `DynamicClassLoader` in their respective classloader hierarchies. The lowest common ancestor of the context classloader for these contexts is an `AppClassLoader`, which is not modifiable. Hence, any modification to the highest modifiable classloader in the first context is not visible in the second.

At load time:

```
;; jdk-sources
#object[java.net.URL 0x2199cb1b file:/usr/lib/jvm/java-11-openjdk/lib/src.zip]

;; classpath
(;; ...all identical to below, except this entry is present:
 #object[java.net.URL 0x2199cb1b "file:/usr/lib/jvm/java-11-openjdk/lib/src.zip"])

;; classloaders
(#object[clojure.lang.DynamicClassLoader 0x6d6bc158 "clojure.lang.DynamicClassLoader@6d6bc158"]
[...]

PR #114, opened 2021-04-11
Add ability to skip dynapath-based functionality

vemv:
Fixes https://github.com/clojure-emacs/orchard/issues/113
Closes https://github.com/clojure-emacs/orchard/pull/112

The disabling is opt-in, i.e. nothing is changed (relative to current behavior) by default.

In order to exercise the proposed changes properly (and also, to prove that a plugin-based approach can work), it is proposed that the build matrix exercises both the dynapath and no-dynapath choices.

⚠️ Between and a tenative addition of JDK16 to the matrix, the matrix would become quite extensive - perhaps a bit over the top?

Green build in my account:

image

⚠️ for the no-dynapath choice _and_ JDK 8, a CI job will take 3m instead of 30-60s. Seems a legit issue, can investigate.

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Changes
 
+* [#113](https://github.com/clojure-emacs/orchard/issues/113) Add ability to skip functionality that works by altering the classpath
+  * You an opt in to this choice by setting `"-Dorchard.use-dynapath=false"`.
 * The _class info cache_ is now initialized silently by default. This results in less confusing output.
   * You can now `@orchard.java/cache-initializer` for deterministically waiting for this cache workload to complete.
   * You can control its verbosity by setting `"-Dorchard.initialize-cache.silent=false"` (or `[...]=true`).
--- a/README.md
+++ b/README.md
@@ -91,8 +91,13 @@ functionality that's provided.
 So far, Orchard follows these options, which can be specified as Java system properties
 (which means that end users can choose to set them globally without fiddling with tooling internals):
 
+* `"-Dorchard.use-dynapath=false"` (default: true)
+  * if `false`, all features that currently depend on dynapath (and therefore alter the classpath) will be disabled.
+  * This is a way to avoid a number of known issues: [#103](https://github.com/clojure-emacs/orchard/issues/103), [#105](https://github.com/clojure-emacs/orchard/issues/105), [#112](https://github.com/clojure-emacs/orchard/pull/112).
+  * Note that if this option is `false`, Orchard clients will have to figure out themselves a way to e.g. fetch Java sources.
+    * It is foreseen that soon enough this will be reliably offered as a Lein plugin.
 * `"-Dorchard.initialize-cache.silent=true"` (default: `true`)
-  * if false, the _class info cache_ initialization may print warnings (possibly spurious ones).
+  * if `false`, the _class info cache_ initialization may print warnings (possibly spurious ones).xx
 
 ## History
 
--- a/project.clj
+++ b/project.clj
@@ -1,3 +1,61 @@
+;;;; The following code allows to add the JDK sources without `dynapath` being present.
+
+(require '[clojure.java.io :as io])
+
+(import '[java.util.zip ZipInputStream]
+        '[java.io FileOutputStream])
+
+(defmacro while-let [[sym expr] & body]
+  `(loop [~sym ~expr]
+     (when ~sym
+       ~@body
+       (recur ~expr))))
+
+(defn jdk-find [f]
+  (let [home (io/file (System/getProperty "java.home"))
+        parent (.getParentFile home)
+        paths [(io/file home f)
+               (io/file home "lib" f)
+               (io/file parent f)
+               (io/file parent "lib" f)]]
+    (->> paths (filter #(.canRead ^java.io.File %)) first str)))
+
+(def jdk-sources
+  (let [java-path->zip-path (fn [path]
+                              (some-> (io/resource path)
+                                      ^java.net.JarURLConnection (. openConnection)
+                                      (. getJarFileURL)
+                                      io/as-file
+                                      str))]
+    (or (java-path->zip-path "java.base/java/lang/Object.java") ; JDK9+
+        (java-path->zip-path "java/lang/Object.java")           ; JDK8-
+        (jdk-find "src.zip"))))
+
+(defn uncompress [path target]
+  (let [zis (-> target io/input-stream ZipInputStream.)]
+    (while-let [entry (-> zis .getNextEntry)]
+      (let [size (-> entry .getSize)
+            bytes (byte-array 1024)
+            dest (->> entry .getName (io/file path))
+            dir (-> entry .getName (clojure.string/split #"/") butlast)
+            _ (->> (clojure.string/join "/" dir) (java.io.File. path) .mkdirs)
+            output (FileOutputStream. dest)]
+        (loop [len (-> zis (.read bytes))]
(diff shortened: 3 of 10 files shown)


project.clj
```diff
@@ -1,3 +1,58 @@
...
+(defn jdk-find [f]
+  (let [home (clojure.java.io/file (System/getProperty "java.home"))
+        parent (.getParentFile home)
+        paths [(clojure.java.io/file home f)
+               (clojure.java.io/file home "lib" f)
+               (clojure.java.io/file parent f)
+               (clojure.java.io/file parent "lib" f)]]
+    (->> paths (filter #(.canRead ^java.io.File %)) first str)))
+
+(def jdk-sources
```
vemv, line 21: These top-level defs can be considered an "inline Lein plugin" for fetching the JDK sources and adding them to the classpath without altering classloaders.

They are not essential to this PR, but they allow the tests depending on the JDK sources being reachable to still pass.

...One could opt instead for disabling the tests when dynapath is disabled, but disabling tests seems delicate (less coverage, less confidence).

It also seems a useful point to show that consumers can add JDK sources _now_, with little code.

Admittedly it would be more tidy to extract this to a Lein plugin, but first I'd like to know what you think of the approach to begin with.

project.clj
```diff
@@ -1,3 +1,58 @@
...
+                   (some-> (clojure.java.io/resource path)
+                           ^java.net.JarURLConnection (. openConnection)
+                                      (. getJarFileURL)
+                                      clojure.java.io/as-file
+                                      str))]
+    (or (java-path->zip-path "java.base/java/lang/Object.java") ; JDK9+
+                   (java-path->zip-path "java/lang/Object.java") ; JDK8-
+        (jdk-find "src.zip"))))
+
+(defn uncompress [path target]
```
vemv, line 32: For some reason simply adding `src.zip` to the classpath didn't work. I tried different variations (prepending `file:`, appending `!/`), no luck.

One has to uncompress src.zip instead, and add the resulting dir to the classpath.

(this adds ~5s to the build, doesn't seem worth caching... I'd be afraid of a given JDK creating a cache for a different JDK)

project.clj
```diff
@@ -38,10 +95,17 @@
...
-             :test {:resource-paths ["test-resources"]
+             :test {:dependencies [[org.clojure/java.classpath "1.0.0"]]
+                    :resource-paths ["test-resources"]
                     ;; Initialize the cache verbosely, as usual, so that possible issues can be more easily diagnosed:
                     :jvm-opts ["-Dorchard.initialize-cache.silent=false"]}
 
+             :no-dynapath {:jvm-opts ["-Dorchard.fetch-java-sources-via-dynapath=false"]
+                           :resource-paths [~(unzipped-jdk-source)]
+                           :plugins ~(if jdk8?
+                                       '[[lein-jdk-tools "0.1.1"]]
```
vemv, line 106: Ref: https://github.com/pallet/lein-jdk-tools

Turned out to be necessary.

README.md
```diff
@@ -91,8 +91,13 @@ functionality that's provided.
 So far, Orchard follows these options, which can be specified as Java system properties
 (which means that end users can choose to set them globally without fiddling with tooling internals):
 
+* `"-Dorchard.fetch-java-sources-via-dynapath=false"` (default: true)
+  * if `false`, all features that currently depend on dynapath (and therefore alter the classpath) will be disabled.
+  * This is a way to avoid a number of known issues.
```
bbatsov, line 96: I guess you can link to #103 and friends here, so this is more specific.

project.clj
```diff
@@ -1,3 +1,58 @@
+(require 'clojure.java.io)
```
bbatsov, line 1: I'd add some section comments around this, explaining the need for it, so casual readers of the project's code are not surprised.

project.clj
bbatsov: It seems reasonable overall. The code itself is faily simple.

project.clj
bbatsov: I think that's fine.

project.clj
bbatsov: Why so?

src/orchard/java.clj
```diff
@@ -58,17 +58,24 @@
                (io/file parent "lib" f)]]
     (->> paths (filter #(.canRead ^File %)) first io/as-url)))
 
+(def fetch-java-sources-via-dynapath?
```
bbatsov, line 61: I think "fetch" is not the right word here, as we're just adding them to the classpath if they are already present locally. Probably "add/load" would be better.

bbatsov:
> ⚠️ Between and a tenative addition of JDK16 to the matrix, the matrix would become quite extensive - perhaps a bit over the top?

Yeah, it'ss nice you simplified it a bit. In general only 8, 11 and the last JDK are needed there. See also https://docs.cider.mx/cider/about/compatibility.html

Overall the PR seems reasonable to me. I guess we should add some instructions about your plugin to CIDER's docs as that's where it's most likelyt for the users to find those.

project.clj
vemv: > Why so?

It compensates, when needed, for the loss of this functionality: https://github.com/clojure-emacs/orchard/blob/f763175a4e8f7b8d7e6df5be1472b3fd9c6ec757/src/orchard/java.clj#L73-L81

vemv (commented):
> In general only 8, 11 and the last JDK are needed there. See also https://docs.cider.mx/cider/about/compatibility.html

👍 will remove jdk15 (for jdk16) then

> I guess we should add some instructions about your plugin to CIDER's docs as that's where it's most likelyt for the users to find those.

If you don't mind, I'd like to defer that a couple weeks: that gives me time to:

* extract the plugin to its own repo
* expand the plugin to also fetch JDK sources when needed
* expand the plugin to also fetch `tools.jar` when needed.

This way, end users will ever be pointed _once_ to a plugin that solves all problems at the same time.

Else it could be confusing to e.g. eventually require (and communicate the need for) N separate plugins, or tell people "you need to upgrade the plugin".

https://github.com/threatgrid/clj-experiments/tree/master/resolve-java-sources-and-javadocs is in a really mature state anyway so I expect the additions to be ready (+ tested/CId) before a couple weeks.

project.clj
bbatsov: Got it.

bbatsov:
> If you don't mind, I'd like to defer that a couple weeks: that gives me time to:

Yeah, totally. I meant this is "we should add them there eventually". :-) I'll cut a new Orchard release once this PR is merged and I'll do another one once we remove the fallback and your plugin is ready for the limelight.

vemv:
Rebased against master, and amended the second commit to take into account PR feedback.

JDK15 has been removed from the CI matrix (for 16)

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Changes
 
+* [#113](https://github.com/clojure-emacs/orchard/issues/113) Add ability to skip functionality that works by altering the classpath
+  * You an opt in to this choice by setting `"-Dorchard.fetch-java-sources-via-dynapath=false"`.
```
bbatsov, line 12: Btw, why don't we just shorted this to `orchard.use-dynapath`? After all the sources were its only use anyways.

vemv: > Btw, why don't we just shorted this to orchard.use-dynapath? After all the sources were its only use anyways.
 
Good call! Done

bbatsov:
I've cut orchard 0.7, and I'm in the process of cutting cider-nrepl 0.26-snapshot

PR #115, opened 2021-04-11
Truncate maps in inspector according to config

yuhan0:
Extends #111 to make maps truncate the same way as other collections, 
showing `*max-coll-size*` number of elements if they are too long.

----

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -205,6 +205,7 @@
     (atom? value)                                  :atom
     (and (instance? Seqable value) (empty? value)) :seq-empty
     (and (map? value) (short? value))              :map
+    (map-entry? value)                             :map-entry
     (map? value)                                   :map-long
     (and (vector? value) (short? value))           :vector
     (vector? value)                                :vector-long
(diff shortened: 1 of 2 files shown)


yuhan0:
I added a missing comma before the `...` to match the previous representation, and changed the :array-long implementation for consistency.
`{ :a 1, :b 2 ... }`
vs 
`{ :a 1, :b 2, ... }`

If preferred I can squash the commits.

bbatsov:
I'm fine with the proposed change, but it needs a  changelog entry. 

I'm also a bit puzzled what's currently the criteria to use commas as separators in some collections and not in others. Seems right now it's just for maps, right? Perhaps we can expose some configuration for this down the road if people want commas for everything sequential.

yuhan0:
Ok, I didn't think it needed a changelog entry as it was an extension to #111.

I just followed the existing behavior with placing commas in maps and arrays. the second commit was to correct a mistake and restore the trailing comma before `...`. 

I do think commas helps readability for homogeneous maps - `{1 3 5 8 2 4 7 3 10 9}` vs `{1 3, 5 8, 2 4, 7 3, 10 9}` 
But arrays don't seem to justify the commas:  `java.lang.Long[] { 0, 1, ... }`. Most likely they were inserted to mimic Java syntax?

bbatsov:
> Ok, I didn't think it needed a changelog entry as it was an extension to #111.

I thought the second commit was fixing a pre-existing bug, but I might have misunderstood that. Anyways, it's something small, so I'm fine if we don't document this. 

> But arrays don't seem to justify the commas: java.lang.Long[] { 0, 1, ... }. Most likely they were inserted to mimic Java syntax?

Perhaps. The code is truly ancient there. :-) We can leave this as is for now.

PR #116, opened 2021-04-16
`orchard.classpath`: don't fail on directories ending on .jar

vemv:
I recently hit the corner case of a directory named `foo.jar` which caused orchard to choke on it.

This PR proposes a fix.

- [x] All tests are passing

### DIFF
--- a/src/orchard/misc.clj
+++ b/src/orchard/misc.clj
@@ -38,12 +38,15 @@
   file extensions"
   [f & exts]
   (when-let [file (if (url? f)
-                    (when  (= (.getProtocol ^java.net.URL f) "file")
+                    (when (= (.getProtocol ^java.net.URL f) "file")
                       (io/as-file f))
                     (io/as-file f))]
-    (some (fn [ext]
-            (.endsWith (.. file getName toLowerCase) ext))
-          exts)))
+    (and
+     ;; Check for file-ness because having a specific extension implies we assume `f` is a file:
+     (not (directory? f))
+     (some (fn [ext]
+             (.endsWith (.. file getName toLowerCase) ext))
+           exts))))
 
 (defn clj-file?  [f] (file-ext? f ".clj" ".cljc"))
 (defn java-file? [f] (file-ext? f ".java"))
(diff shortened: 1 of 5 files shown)


test/orchard/java/classpath_test.clj
```diff
@@ -22,22 +22,29 @@
 (deftest classpath-test
   (testing "Classpath"
     (testing "URLs are absolute file paths"
-      (is (every? #(.isAbsolute (File. (.getPath %)))
```
vemv, line 25: Changed all instances of `is every?` to `is` inside a `doseq`. 

This way:

* the precise thing that failed is reported
* we avoid huge failure reports, especially now that we added the `"unzipped-jdk-source"` resource path

src/orchard/misc.clj
```diff
@@ -41,9 +41,11 @@
                     (when  (= (.getProtocol ^java.net.URL f) "file")
                       (io/as-file f))
                     (io/as-file f))]
-    (some (fn [ext]
-            (.endsWith (.. file getName toLowerCase) ext))
-          exts)))
+    (and (some (fn [ext]
```
bbatsov, line 44: Doesn't it make more sense to swap the order of changes - check first that something is not a folder? That reads better to me, and probably won't have much impact on performance.

test/orchard/java/classpath_test.clj
bbatsov: Agreed!

bbatsov:
This needs a changelog entry and we're good to go.

bbatsov:
I pressed "Close" instead of "Comment". Sorry!

test/orchard/java/classpath_test.clj
bbatsov: Btw, I'd even suggest adding this as a guideline to the `Clojure Style Guide` under "Testing".

vemv:
> Btw, I'd even suggest adding this as a guideline to the Clojure Style Guide under "Testing".

:) yeah it's something that pops up often in code review. Might contribute at some point.

Feedback applied as a git `amend`

bbatsov:
Thanks!

vemv:
🍻!

Would appreciate a release if possible - this was a nuisance while hacking on refactor-nrepl (I had that .jar dir auto-generated for some unrelated reason, causing an error in refactor-nrepl code)

bbatsov:
Sure, I'll cut 0.7.1 as soon as I can.

bbatsov:
The new release is out.

PR #119, opened 2021-08-10
Parallelize `orchard.namespace`

vemv:
Parallelizes `orchard.namespace` when it makes sense (namely, IO-heavy ops) and it's safe to do so (i.e., no code loading performed).

This matters particularly in projects with numerous namespaces and large classpaths.

Makes https://github.com/clojure-emacs/cider-nrepl/pull/701 more efficient, particularly when `(namespace/project-namespaces)` will be invoked slightly more often (since a `def` became a `defn`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* `orchard.namespace` functionality is now parallelized when possible.
+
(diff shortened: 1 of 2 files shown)


bbatsov:
The changes look good, just add some notes/comments about the need for the parallelization, so people are not surprised by it.

vemv:
Cheers, added a ns docstring.

PR #123, opened 2021-09-07
Fix info lookups from namespaces that don't yet exist

plexus:
Currently when doing a `var-info` from a file before evaluating the `ns` form
always fails. The namespace object does not yet exist, and so none of the var
resolution strategies work.

However fully-qualified vars, namespace names, or built-ins (clojure.core
functions) could all be resolved, and it's frustrating that they aren't.

This adds a check to see if `(find-ns ns)` returns anything. If not then we are
initiating the lookup from a namespace that has not yet been defined, in which
case none of the lookup strategies will proceed, so we continue with doing the
lookup as if it had been initiated from `clojure.core`.

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/orchard/info.clj
+++ b/src/orchard/info.clj
@@ -63,30 +63,34 @@
 
 (defn clj-meta
   {:added "0.5"}
-  [{:keys [dialect ns sym computed-ns unqualified-sym]}]
+  [{:keys [dialect ns sym computed-ns unqualified-sym] :as opts}]
   {:pre [(= dialect :clj)]}
   (let [ns (or ns computed-ns)]
-    (or
-     ;; it's a special (special-symbol?)
-     (m/special-sym-meta sym)
-     ;; it's a var
-     (some-> ns (m/resolve-var sym) (m/var-meta))
-     ;; it's a Java constructor/static member symbol
-     (some-> ns (java/resolve-symbol sym))
-     ;; it's an unqualified sym maybe referred
-     (some-> ns (m/resolve-var unqualified-sym) (m/var-meta))
-     ;; it's a Java class/record type symbol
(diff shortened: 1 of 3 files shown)


src/orchard/info.clj
```diff
@@ -63,30 +63,34 @@
...
+       ;; too ambiguous.
+       ;;
+       ;; Observe the incorrect behavior (should return nil, there is a test):
+       ;;
+       ;;   (info '{:ns clojure.core :sym non-existing}) ;;=> {:author "Rich Hickey" :ns clojure.core ...}
+       ;;
+       (some-> (find-ns unqualified-sym) (m/ns-meta)))
+      ;; Lookups in files whose namespaces don't exist yet should still be able
+      ;; to resolve built-ins and fully-qualified syms
+      (recur (assoc opts :ns 'clojure.core)))))
```
vemv, line 93: Very minor, but the large distance between the `if` and this branch makes it a bit hard to track the control flow.

By using an `if-not` instead one gets these "degenerate cases" out of one's way, similarly to https://wiki.c2.com/?GuardClause

vemv (commented):
LGTM, the idea seems sound. Probably trying to `require` things would be excessively obstrusive.

Maybe worth adding some test coverage.

src/orchard/info.clj
plexus: Thanks for the feedback. I've addressed this one. I'll still be adding a test and changelog etc.

plexus:
Added a test, appended the CHANGELOG, and swapped the if branches around as suggested by @vemv . I think from my side this is good to go.

vemv (approved):
🍻

bbatsov:
Nice improvement. Thanks!

PR #124, opened 2021-09-18
Retain resource URL in java-parser results, improve caching

plexus:
The java parser/legacy-parser returns a resource's `:file`, actually a relative
path as a string, and `:path`, the relative path converted to absolute. It does
this conversion through an `io/resource` lookup to find the file system location
of the artifact. There are two issues with this:

1. The return value of `io/resource` is discarded, even though it is needed
later on in `orchard.java`, causing unnecessary resource lookups.

2. The path is obtained through `(.getPath (io/resource path))`, which yields
nonsensical results when used on resources that are inside a JAR.

This change keeps the `:file` returned from java-parser (relative path string),
but removes the `:path` value in favor of a `:resource-url`, i.e. the return
value of `(io/resource path)`. It then provides utility functions to work with
this resource URL directly, namely mapping it to a filesystem artifact, or
retrieving its modification time.

java.parser/parser already does a `io/resource` lookup for the resource it is
parsing, meaning it has a full URL (jar: or file:). By including this URL in the
map it returns callers can do further checks or operations on the resource
without having to re-scan the classpath.

This in turn allows us to simplify and optimize `orchard.java/class-info`. The
old version would call `io/resource` on each invocation, even when the result
was already cached, in order to find the artifact's modification time.

This can noticably speed up cider-nrepl's `"stacktrace"` op, which calls
`orchard.java/class-info` on each stack frame, in extreme cases taking multiple
seconds to analyze all stack frames and return a result. This leads to rather
jarring UX, with the stacktrace buffering popping up in Emacs seconds after the
evaluation and exception happened. This is exarcerbated when using the nREPL
sideloader, which adds overhead to each resource lookup.

This also makes the return value of java-parser more correct. As mentioned the
previous version would always call `(.getPath (io/resource path))`, but this
only makes sense for `file:` resources, not for `jar:` resources. For `file:`
URLs `.getPath` returns the path of the file. For `jar:` URLs it returns the
nested url+the jar entry, so a `file:` URL but with a dangling `!/jar/entry`.

Illustration of why calling `getPath` on `jar:` URLs is not useful:

```clojure
(io/resource "lambdaisland/witchcraft.clj")
;;=> #java.net.URL "file:/srv/mc/witchcraft/src/lambdaisland/witchcraft.clj"

(.getPath (io/resource "lambdaisland/witchcraft.clj"))
;;=> "/srv/mc/witchcraft/src/lambdaisland/witchcraft.clj"

(io/resource "clojure/lang/RT.class")
;;=> #java.net.URL "jar:file:/root/.m2/repository/org/clojure/clojure/1.10.3/clojure-1.10.3.jar!/clojure/lang/RT.class"

(.getPath (io/resource "clojure/lang/RT.class"))
;;=> "file:/root/.m2/repository/org/clojure/clojure/1.10.3/clojure-1.10.3.jar!/clojure/lang/RT.class"
```

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -1,4 +1,4 @@
 {:output {:progress true
           :exclude-files ["analysis.cljc" "meta.cljc" "inspect_test.clj"]}
- :linters {:unused-private-var {:level :warning :exclude [orchard.query-test/a-private orchard.query-test/docd-fn]}}}
-
+ :linters {:unused-private-var {:level :warning :exclude [orchard.query-test/a-private orchard.query-test/docd-fn]}
+           :refer-all {:exclude [clojure.test]}}}
--- a/src/orchard/info.clj
+++ b/src/orchard/info.clj
@@ -195,3 +195,7 @@
   classes. If no source is available, return the relative path as is."
   [^String path]
   {:javadoc (java/resolve-javadoc-path path)})
+
+(comment
+  (info-java 'clojure.lang.RT 'baseLoader)
+  (file-info "clojure/core.clj"))
--- a/src/orchard/util/io.clj
+++ b/src/orchard/util/io.clj
@@ -1,4 +1,10 @@
-(ns orchard.util.io)
+(ns orchard.util.io
+  "Utility functions for dealing with file system objects, and in/out streams."
+  (:require
+   [clojure.java.io :as io])
+  (:import
+   (java.io File)
+   (java.net URL JarURLConnection)))
 
 (defn wrap-silently
   "Middleware that executes `(f)` without printing to `System/out` or `System/err`.
@@ -22,3 +28,55 @@
             (System/setOut old-out))
           (when (= ps System/err) ;; `System/err` may have changed in the meantime (in face of concurrency)
             (System/setErr old-err)))))))
+
+(defn url-protocol
+  "Get the URL protocol as a string, e.g. http, file, jar."
+  [^java.net.URL url]
+  (.getProtocol url))
+
+(defn url-to-file-within-archive?
+  "Does this URL point to a file inside a jar (or zip) archive.
+  i.e. does it use the jar: protocol."
+  [^java.net.URL url]
+  (= "jar" (url-protocol url)))
+
+(defn resource-jarfile
+  "Given a jar:file:...!/... URL, return the location of the jar file on the
+  filesystem. Returns nil on any other URL."
+  ^File [^URL jar-resource]
+  (assert (= "jar" (url-protocol jar-resource)))
+  (let [^JarURLConnection conn (.openConnection jar-resource)
+        inner-url (.getJarFileURL conn)]
+    (when (= "file" (url-protocol inner-url))
+      (io/as-file inner-url))))
+
+(defn resource-artifact
+  "Return the File from which the given resource URL would be loaded.
+
+  For `file:` URLs returns the location of the resource itself, for
+  `jar:..!/...` URLs returns the location of the archive containing the
+  resource. Returns a fully qualified File, even when the URL is relative.
+  Throws when the URL is not a `file:` or `jar:` URL."
+  ^File [^java.net.URL resource]
+  (let [protocol (url-protocol resource)]
+    (case protocol
+      "file"
+      (io/as-file resource)
+      "jar"
+      (resource-jarfile resource)
+      #_else
+      (throw (ex-info (str "URLs with a " protocol
+                           " protocol can't be situated on the filesystem.")
+                      {:resource resource})))))
+
+(defprotocol LastModifiedTime
+  (last-modified-time [this]
+    "Return the last modified time of a File or resource URL."))
+
+(extend-protocol LastModifiedTime
+  java.net.URL
+  (last-modified-time [this]
+    (last-modified-time (resource-artifact this)))
+  java.io.File
+  (last-modified-time [this]
+    (.lastModified this)))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#123](https://github.com/clojure-emacs/orchard/pull/123): Fix info lookups from namespaces that don't yet exist
 * [#125](https://github.com/clojure-emacs/orchard/issues/125): Don't fail if the classpath references a non-existing .jar
 * [#128](https://github.com/clojure-emacs/orchard/issues/128): Strengthen `apropos`
+* [#124](https://github.com/clojure-emacs/orchard/pull/124): Remove costly `io/resource` lookup
 
 ## 0.7.1 (2021-04-18)
 
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -14,5 +14,6 @@
 
 (cond->> ["dev" "src" "test"]
   jdk8?       (into ["src-jdk8"])
-  (not jdk8?) (into ["src-newer-jdks"])
+  (not jdk8?) (into ["src-newer-jdks"
+                     "test-newer-jdks"])
   true        (apply set-refresh-dirs))
--- a/project.clj
+++ b/project.clj
@@ -87,6 +87,9 @@
              "-Dclojure.main.report=stderr"]
 
   :source-paths ["src" "src-jdk8" "src-newer-jdks"]
+  :test-paths ~(cond-> ["test"]
+                 (not jdk8?)
+                 (conj "test-newer-jdks"))
 
   :profiles {
              ;; Clojure versions matrix
@@ -96,8 +99,8 @@
                                   [org.clojure/clojure "1.8.0" :classifier "sources"]]}
              :1.9 {:dependencies [[org.clojure/clojure "1.9.0"]
                                   [org.clojure/clojure "1.9.0" :classifier "sources"]]}
-             :1.10 {:dependencies [[org.clojure/clojure "1.10.1"]
-                                   [org.clojure/clojure "1.10.1" :classifier "sources"]]}
+             :1.10 {:dependencies [[org.clojure/clojure "1.10.3"]
+                                   [org.clojure/clojure "1.10.3" :classifier "sources"]]}
              :master {:repositories [["snapshots"
                                       "https://oss.sonatype.org/content/repositories/snapshots"]]
                       :dependencies [[org.clojure/clojure "1.11.0-master-SNAPSHOT"]
--- a/src-jdk8/orchard/java/legacy_parser.clj
+++ b/src-jdk8/orchard/java/legacy_parser.clj
@@ -273,6 +273,8 @@
         (assoc (->> (map parse-info (.classes root))
                     (filter #(= klass (:class %)))
                     (first))
+               ;; relative path on the classpath
                :file path
-               :path (. (io/resource path) getPath))))
+               ;; Full URL, e.g. file:.. or jar:...
+               :resource-url (io/resource path))))
     (catch Abort _)))
--- a/src-newer-jdks/orchard/java/parser.clj
+++ b/src-newer-jdks/orchard/java/parser.clj
@@ -304,7 +304,9 @@
                       (map #(parse-info % root))
                       (filter #(= klass (:class %)))
                       (first))
+                 ;; relative path on the classpath
                  :file path
-                 :path (.getPath (io/resource path)))
+                 ;; Full URL, e.g. file:.. or jar:...
+                 :resource-url (io/resource path))
           (finally (.close (.getJavaFileManager root))))))
     (catch Throwable _)))
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -250,16 +250,6 @@
 
 (def cache (atom {}))
 
-(defn- immutable-source-file?
-  "Return true if the source file is effectively immutable. Specifically, this
-  returns true if no source file is available, or if the source file is in a
-  jar/zip archive."
-  [info]
-  (let [path (:file info)
-        src  (when path (io/resource path))]
-    (or (not src)
-        (re-find #"\.(jar|zip)!" (str src)))))
-
 (defn class-info
   "For the class symbol, return (possibly cached) Java class and member info.
   Members are indexed first by name, and then by argument types to list all
@@ -269,9 +259,11 @@
         info (if cached
                (:info cached)
                (class-info* class))
-        last-modified (if (immutable-source-file? info)
+        resource (:resource-url info)
+        last-modified (if (or (nil? resource)
+                              (util.io/url-to-file-within-archive? resource))
                         0
-                        (.lastModified ^File (io/file (:path info))))
+                        (util.io/last-modified-time resource))
         stale (not= last-modified (:last-modified cached))
         ;; If last-modified in cache mismatches last-modified of the file,
         ;; regenerate class-info.
--- a/src/orchard/java/resource.clj
+++ b/src/orchard/java/resource.clj
@@ -15,7 +15,9 @@
     s))
 
 (defn project-resources
-  "Get a list of classpath resources."
+  "Get a list of classpath resources, i.e. files that are not clojure/java source
+  or class files. Only consider classpath entries that are directories, does not
+  consider jars."
   []
   (mapcat
    (fn [^File directory]
@@ -32,7 +34,7 @@
                    {:root directory
                     :file file
                     :relpath relpath
-                    :url (io/resource relpath)})))
+                    :url (io/as-url file)})))
           (remove #(.startsWith ^String (:relpath %) "META-INF/"))
           (remove #(re-matches #".*\.(clj[cs]?|java|class)" (:relpath %)))))
    (filter (memfn ^File isDirectory) (map io/as-file (cp/classpath (cp/boot-aware-classloader))))))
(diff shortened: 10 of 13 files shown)


vemv (commented):
Awesome! Appreciate the PR desc.

I guess this could have test coverage both directly, and more importantly it's worth checking out the consumer of this defn - surely it lacked test coverage as well.

bbatsov:
Great PR indeed! If we can easily improve the test coverage, that'd be great, but I'm willing to take the improvement regardless.

plexus:
The consumer has tests, but no tests that consider `:lastModified`. It's a bit of a pain to test, because you'd need a `class` file that is not inside a jar. I guess you could bundle a pre-compiled `.class` just for testing, or include a Java file that you compile on the fly.

Not sure if tests for `immutable-source-file?` would have helped, since it seems the issue is assumptions made about its inputs. You'd only really notice if you had tests that combine the java parser and `immutable-source-file?`. But I can add a few tests that demonstrate the input it gets, and maybe add a comment or two.

The current behavior is that `immutable-source-file?` always returns true, (because the `io/resource` call always returns `nil`) which means it never expires the cached info about classes. Which for most users will actually be fine, unless they're using something like Virgil to work on Java and Clojure simultaneously. But despite the caching the `immutable-source-file?` check still happens every time (to decide if it should expire), and that check was needlessly costly.

Actually we could go a step further here and skip the check on class info that is already cached and coming from a JAR, on the assumption that if a class is coming from a JAR now it will still be coming from a JAR in the future.

Seems this also needs a rebase now, I'll get to that.

bbatsov:
> Actually we could go a step further here and skip the check on class info that is already cached and coming from a JAR, on the assumption that if a class is coming from a JAR now it will still be coming from a JAR in the future.

I like this idea! 

> The consumer has tests, but no tests that consider :lastModified. It's a bit of a pain to test, because you'd need a class file that is not inside a jar. I guess you could bundle a pre-compiled .class just for testing, or include a Java file that you compile on the fly.

That's too much hassle for little return, so let's not bother with it.

plexus:
Looking at this a bit closer the problem is really in java-parser

```clojure
          (assoc ...
                 :file path
                 :path (.getPath (io/resource path)))
```

`io/resource` returns a `URL`. If that's a `file:` URL then `.getPath` will indeed return the path on the filesystem, that seems to be mostly what this code is assuming. But it could also be a `jar:` or even an `http:`, `ftp:` or `data:` URL (depending on how funky your classloader is). The main cases to consider here are `file:` and `jar:` URL, but `.getPath` on a `jar:` URL pretty much gives you gibberish, because it strips the `jar:` but leaves the `!/...` at the end. (jar URLs look like `jar:<url>/!<entry>`, and so you get `<url>/!<entry>`). The correct way to parse a jar URL is apparently via [`JarURLConnection`](https://docs.oracle.com/javase/7/docs/api/java/net/JarURLConnection.html)

Also keep in mind that the `<url>` inside the `jar:<url>..` doesn't have to be a `file:` url, it could also e.g. be `http:`.

So I think the better solution is actually for java-parser to do

```
(assoc ... :file path :resource-url (io/resource path))
```

and skip the `.getPath` alltogether.

If it's possible that other callers might still be relying on `:path` we can leave it in the result as well.

plexus:
Glad I slept on this, on closer inspection `immutable-source-file?` did return the right result, but in a way that unnecessarily calls `io/resource` even though this is already done in `source-info`, and in a way that bypasses the cache check.

I've changed the behavior of `parser/source-file` slightly, for the case where a file is coming from a jar

```clojure
;; Before
(select-keys (parser/source-info 'clojure.lang.RT) [:file :path :archive?])
{:file "clojure/lang/RT.java", :path "file:/home/arne/.m2/repository/org/clojure/clojure/1.10.1/clojure-1.10.1-sources.jar!/clojure/lang/RT.java"}

;; After
(select-keys (parser/source-info 'clojure.lang.RT) [:file :path :archive?])
{:file "clojure/lang/RT.java", :path "/home/arne/.m2/repository/org/clojure/clojure/1.10.1/clojure-1.10.1-sources.jar", :archive? true}
```

So that `:path` is always a filesystem path, just as it is with non-archive resources. I've also added the `:archive?` key so we don't have to scan for `.(jar|zip)!/` later on.

```
;; non-archive resources
(select-keys (parser/source-info 'orchard.java.DummyClass) [:file :path :archive?])
{:file "orchard/java/DummyClass.java", :path "/home/arne/github/orchard/test/orchard/java/DummyClass.java", :archive? false}
```

plexus:
I can't immediately find anywhere in CIDER where `:path` is used, only `:file`.

Do you have any pointers for things to check? This is specifically for java class info, so things like `find-var` or `find-ns` should not be impacted. I ended up here because of its use in `stacktrace`, and there `cider-stacktrace` is definitely only checking `:file`.

The thing is that when `info` ops return `:file` has been made absolute, if you look at cider.nrepl.middleware.info.

``` clojure
      (-> info
          (merge
               (when-let [file (:file info)]
                   (info/file-info file))
)
)
```

Where `info/info-file` does another resource lookup, this logic is also somewhat flawed

``` clojure
(defn file-info
  [path]
  (let [[resource-relative resource-full] (resource/resource-path-tuple path)]
    (merge {:file (or (file-path path) resource-full path)}
           ;; Classpath-relative path if possible
           (when resource-relative
             {:resource resource-relative}))))
```

Is this path a classpath-relative path, or a CWD-relative path? `resource-path-tuple` assumes the former, `file-path` assumes the latter.

And in this case (when the info is coming from java.parser) we have all this information already...

And then in `cider-stacktrace-navigate` this is basically ignored, instead it tries to guess first if it can figure out itself where that class source is, even if we just told it where it is...

```emacs-lisp
(let (
         (file (or
                (and (null var) (cider-resolve-java-class class))
                (nrepl-dict-get info "file")
                (button-get button 'file)))
)
```

bbatsov:
> Do you have any pointers for things to check? This is specifically for java class info, so things like find-var or find-ns should not be impacted. I ended up here because of its use in stacktrace, and there cider-stacktrace is definitely only checking :file.

I briefly checked the code and it seems you're right. 

> Is this path a classpath-relative path, or a CWD-relative path? resource-path-tuple assumes the former, file-path assumes the latter.

I think all relative paths should be classpath-relative. 

> And then in cider-stacktrace-navigate this is basically ignored, instead it tries to guess first if it can figure out itself where that class source is, even if we just told it where it is...

Too much legacy - I'm pretty sure we've slowly accumulated a lot of small inconsistencies over the years. I agree the current behavior is flawed.

plexus:
ok, in that case let's take this from the source to the consumer step by step, and try to be consistent about the meaning of each key. In the end there are two things that matter: the classpath-relative path, and the full file/jar URL.

The info maps coming from orchard seem to all use `:file` as the classpath-relative path. However in the `info` op in cider-nrepl `:file` is (at least most of the time it seems?) a full URL (file: or jar:) which makes sense. Except if orchard gives it a relative path that it can't resolve, in that case it stays a relative path...

For your own REPL perusal

```clojure
(require '[orchard.info :as info] '[orchard.java :as java] '[cider.nrepl.middleware.info :as cider-info])

(info/info 'clojure.core 'str)
(info/info-java 'clojure.lang.RT 'baseLoader)
(java/class-info 'clojure.lang.RT)

(cider-info/format-response (cider-info/info {:ns 'clojure.core :sym 'str}))
(cider-info/format-response (cider-info/info {:ns 'cider.nrepl.middleware.info :sym 'info}))
(cider-info/format-response (cider-info/info {:ns 'clojure.lang.RT :sym 'baseLoader}))
```

My current thinking is this: we look at the places in `orchard.info` and `orchard.java` that return info-like maps with a `:file` key. If the code creating these maps already has done resource resolution then it adds an extra `:resource-url` as a string (so a `file:` or `jar:` url). And we provide helpers for getting the actual filesystem object from there, e.g. for when you need to know the modification time.

The cider.nrepl.middleware.info can use `:resource-url` as its `:file` in "file" ops, if it is present. This should save a lot of resource lookups already, and yield more consistent results. Same for nrepl.lookup.

Finally we can see how CIDER uses this information, heuristics should only be used when no `:file` is provided.

src/orchard/java/parser.clj
```diff
@@ -285,6 +286,18 @@
         (str module "/" path)
         path))))
 
+(def url-protocol (memfn ^URL getProtocol))
+
+(defn jar-path
+  "Given a jar:file:...!/... URL, return the location of the jar file on the
+  filesystem. Returns nil on any other URL."
+  [^URL jar-resource]
+  (when (= "jar" (url-protocol jar-resource))
```
vemv, line 295:
```clj
(assert (= "jar" (url-protocol jar-resource))
  ;; debug info:
  (url-protocol jar-resource))
```

is just as cheap, but also ensures our assumptions hold true. Specifically, `jar-path` has only one caller, so it's pretty easy to reason about what our preconditions are

vemv (commented):
I'll be interested in reviewing this PR again later, I guess it's a WIP again

(will be nice to have an updated title / desc summarizing the newer findings)

vemv:
One last note, from this branch's CI timings, it looks like certain jobs got slower (30s -> 60s)

(other jobs are known to take 2m+ minutes)

This is worth keeping in mind, ideally we wouldn't regress on performance. I _think_ that if solving all these problems at once we'd get a fast build for every job, see also https://github.com/clojure-emacs/orchard/pull/131#issuecomment-926346748

plexus:
I pushed a new iteration which simply returns the `:resource-url` from the parser, and then provides some helpers in `orchard.misc.io` to work on the resource URL. Going to try this locally with a patched cider-nrepl and report back.

----

Having looked around a bit more it seems it's going to be challenging to standardize the different keys used for things, e.g.

orchard.java.resource
- :root : classpath entry, java.io.File
- :file : resource location, java.io.File
- :relpath : relative resource path, String
- :url : resource location, java.net.URL

orchard.java.parser
- :file : relative resource path, String

orchard.info/file-info
- :file : resource location, java.net.URL
- :resource : relative resource path, String

orchard.info/info*
- :file : relative resource path, String

plexus:
I think this is good from my end. It's a fairly conservative change so far, it'll get more interesting once `cider-nrepl` can starting taking advantage of the `:resource-url` instead of doing another `io/resource` lookup. That said this should already be useful for any Orchard consumer because of the better caching of java results.

src/orchard/util/io.clj
```diff
@@ -22,3 +27,52 @@
             (System/setOut old-out))
           (when (= ps System/err) ;; `System/err` may have changed in the meantime (in face of concurrency)
             (System/setErr old-err)))))))
+
+(defn url-protocol
+  "Get the URL protocol as a string, e.g. http, file, jar."
+  [^java.net.URL url]
+  (.getProtocol url))
+
+(defn in-archive?
```
vemv, line 36: `file-within-archive?` perhaps?

Originally I thought this meant a different thing ("is this an archive")

src/orchard/util/io.clj
```diff
@@ -22,3 +27,52 @@
...
+  "Does this URL point to a file inside a jar (or zip) archive.
+  i.e. does it use the jar: protocol."
+  [^java.net.URL url]
+  (= "jar" (url-protocol url)))
+
+(defn jar-location
```
vemv, line 42: `jar-location-of-file-within-archive`?

More verbose, but:

* consistent w/ my prior suggestion
* does not take a generic name for future use cases

src/orchard/util/io.clj
```diff
@@ -22,3 +27,52 @@
...
+  filesystem. Returns nil on any other URL."
+  ^File [^URL jar-resource]
+  (assert (= "jar" (url-protocol jar-resource)))
+  (let [^JarURLConnection conn (.openConnection jar-resource)
+        inner-url (.getJarFileURL conn)]
+    (when (= "file" (url-protocol inner-url))
+      (io/as-file inner-url))))
+
+(defn resource-artifact
+  "Get the fully qualified file name containing the resource identified by the
```
vemv, line 53: `Get ` is confusing, lead me to believe it referred to the return value at first

test/orchard/java/parser_test.clj
```diff
@@ -0,0 +1,67 @@
+(ns orchard.java.parser-test
+  (:require [orchard.misc :as misc]
+            [clojure.test :refer :all]))
+
+(when (>= orchard.misc/java-api-version 9)
```
vemv, line 5: Can you follow instead this pattern? https://github.com/clojure-emacs/orchard/commit/ffb71cd8935889a57b46a540f4c8b76df667272e

vemv (changes requested):
Is the PR desc up to date?

src/orchard/util/io.clj
bbatsov: I'm fine with a shorter name as the longer seems extremely verbose. I'm all for clear naming, but we don't want to go overboard here.

src/orchard/util/io.clj
bbatsov: Isn't this clear from the param? To me it seems obviously that we're checking if an URL is in archive.

plexus:
Thanks for the input @vemv and @bbatsov . I'll try to make some time to give this another pass.

src/orchard/info.clj
```diff
@@ -198,3 +198,7 @@
   classes. If no source is available, return the relative path as is."
   [^String path]
   {:javadoc (java/resolve-javadoc-path path)})
+
+(comment
```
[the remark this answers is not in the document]
plexus: I was meaning to keep it.

src/orchard/util/io.clj
vemv: > To me it seems obviously that we're checking if an URL is in archive.

Hindsight is always 20/20 :) as mentioned at first the name genuinely confused me. If we pick this code up some years later we can forget these meanings, which is why being extra careful wrt clarity seems desirable for a project like this one.

No biggie though!

plexus:
Rebased, going through the review comments now, and will then update the description.

plexus:
Squashed, addressed the review comments, updated the commit message and PR title/desc.

vemv:
Eyeing in a bit 👀

src/orchard/util/io.clj
plexus: This is now `url-to-file-within-archive?` as in "does this URL point at a file within an archive?"

src/orchard/util/io.clj
plexus: this is now `resource-jarfile`, since it returns the `.jar` `File` instance for a given resource (resources being URLs).

src/orchard/util/io.clj
plexus: Rephrased, thanks.

test/orchard/java/parser_test.clj
plexus: Done, I added `test-newer-jdks` analogous to `src-newer-jdks`

src/orchard/info.clj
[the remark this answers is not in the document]
plexus: Got rid of it after all, the tests provide enough examples.

vemv:
> Illustration of why calling getPath on jar: URLs is not useful:

> (.getPath (io/resource "clojure/lang/RT.class"))
> ;;=> "file:/root/.m2/repository/org/clojure/clojure/1.10.3/clojure-1.10.3.jar!/clojure/lang/RT.class"

The string may not be directly useful but it can be successfully, meaningfully parsed with a simple algo. I think it was (is?) done in cider.el.

I also happen to have Emacs code that parses such strings from cider-nrepl.

Regardless, out of caution we can stick to a simple no-breakage policy

src/orchard/util/io.clj
```diff
@@ -22,3 +28,55 @@
...
+  (= "jar" (url-protocol url)))
+
+(defn resource-jarfile
+  "Given a jar:file:...!/... URL, return the location of the jar file on the
+  filesystem. Returns nil on any other URL."
+  ^File [^URL jar-resource]
+  (assert (= "jar" (url-protocol jar-resource)))
+  (let [^JarURLConnection conn (.openConnection jar-resource)
+        inner-url (.getJarFileURL conn)]
+    (when (= "file" (url-protocol inner-url))
```
vemv, line 50: Seems that this one could have a pred akin to `url-to-file-within-archive?`, for symmetry

.clj-kondo/config.edn
```diff
@@ -1,4 +1,4 @@
 {:output {:progress true
           :exclude-files ["analysis.cljc" "meta.cljc" "inspect_test.clj"]}
- :linters {:unused-private-var {:level :warning :exclude [orchard.query-test/a-private orchard.query-test/docd-fn]}}}
-
+ :linters {:unused-private-var {:level :warning :exclude [orchard.query-test/a-private orchard.query-test/docd-fn]}
+           :refer-all {:exclude [clojure.test]}}}
```
vemv, line 4: We're not using this in other clojure-emacs projects, however tempting

kondo already tells one what should be `:refer`ed, so it's easy enough to comply

vemv (commented):
If you have the chance to QA this (cider.el -> cider-nrepl -> orchard) it would probably be greatly useful. cider.el's test suite is not exactly an e2e one.

This project has a `make install` command (after which, I reckon, you should run cider-nrepl's own `make install`)

vemv:
Can't push to your branch (`permission denied`), can you confirm `Allow edits` is on?

plexus:
Screenshot_20211108-162122

vemv:
Thanks! Yeah it wasn't working for w/e reason 🌀

Issue #90, opened 2020-04-21
cljdoc is failing to compile


mauricioszabo:
The latest (and some older versions) of the documentation is now showing - it's just showing `API import failed in build  #28834`, and it points to https://cljdoc.org/builds/28834

Is it something easy to fix? Or is there any way to build the documentation locally, so we can explore the API?

bbatsov:
The docs fail to compile because cljdoc doesn't work with Java 9+ :( . See my comment on this ticket https://github.com/cljdoc/cljdoc/issues/275

boxxxie:
Is there any working documentation that you guys can point me to? or maybe add a section on the README for how to build the docs myself.

vemv:
We'll see whether it worked with the next release on Clojars

PR #129, opened 2021-09-24
Set up `refresh-dirs` for Orchard development

vemv:
Splits `java.parser` / `java.legacy-parser` into distinct `:source-paths`.

This way, we can conditionally set the `refresh-dirs` for those of us using `tools.namespace` / `cider-refresh`, without it being broken because either of these namespaces couldn't be loaded under one's JDK.


bbatsov:
Looks like a reasonable approach. At some point we should figure out something similar to get cljdoc to build (they use JDK 8, which is why their task explodes for `java.parser`).

bbatsov:
See also https://github.com/clojure-emacs/orchard/issues/90

PR #130, opened 2021-09-24
Try fixing a bug in `apropos`

vemv:
See https://github.com/clojure-emacs/orchard/issues/128

Tries fixing the bug itself, and also introduces a fallback. Will comment inline.

### DIFF
--- a/src/orchard/apropos.clj
+++ b/src/orchard/apropos.clj
@@ -17,28 +17,44 @@
 
 ;;; ## Symbol Search
 
-(defn- apropos-sort
+(defn- safe-comparator [x y]
+  (compare (pr-str x) (pr-str y)))
+
+(defn- default-comparator [ns clojure-ns?]
+  (fn [x y]
+    (cond
+      (= x y)  0
+      (nil? x) 1
+      (nil? y) -1
+      (= x ns) -1
+      (= y ns)  1
+      (and (clojure-ns? x) (not (clojure-ns? y))) -1
+      (and (clojure-ns? y) (not (clojure-ns? x)))  1
+      :else (safe-comparator x y))))
+
+(defn apropos-sort
   "Return a list of vars, ordered with `ns` first,
   followed by `clojure.*` namespaces, and then all others sorted
   alphabetically."
   [ns vars]
-  (let [clojure-ns? #(.startsWith (str (ns-name %)) "clojure.")]
-    (sort-by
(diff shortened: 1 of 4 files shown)


src/orchard/apropos.clj
```diff
@@ -17,28 +17,44 @@
...
 ;;; ## Symbol Search
 
-(defn- apropos-sort
+(defn- safe-comparator [x y]
+  (compare (pr-str x) (pr-str y)))
+
+(defn- default-comparator [ns clojure-ns?]
+  (fn [x y]
+    (cond
+      (= x y)  0
```
vemv, line 26: Added a `(= x y)  0` making the sort stable.

Unfortunately I couldn't prove this fixes the issue, however it's most likely an improvement.

src/orchard/apropos.clj
```diff
@@ -17,28 +17,44 @@
...
+    ;; https://clojure.org/guides/comparators
+    (try
+      (sort-by key-fn (default-comparator ns clojure-ns?) vars)
+      ;; Handle https://github.com/clojure-emacs/orchard/issues/128
+      (catch IllegalArgumentException e
+        (when (System/getProperty "orchard.internal.test-suite-running")
+          ;; Don't accept this exception in our CI - we should fix this if it's reproducible.
+          (throw e))
+        ;; Fallback to a simpler comparator:
+        (sort-by key-fn safe-comparator vars)))))
```
vemv, line 53: Added a fallback so that this feature doesn't fail altogether due to something relatively unimportant such as sorting.

test/orchard/apropos_test.clj
```diff
@@ -3,13 +3,19 @@
...
    [clojure.test :refer [are deftest is testing]]
-   [orchard.apropos :refer [find-symbols]]
+   [orchard.apropos :as sut :refer [find-symbols]]
    [orchard.meta :refer [var-name var-doc]]))
 
 (def ^{:doc "Test1. Test2. Test3."} public-var [1 2 3])
 
 (defn some-random-function [])
 
+(deftest apropos-sort-test
```
vemv, line 13: Added some extra tests, they passed before the fix anyway.

https://github.com/clojure-emacs/orchard/issues/128 seems hard to repro since apropos results depend on ones's classpath.

bbatsov:
The proposed fix seems good to me.

PR #131, opened 2021-09-24
CI improvements

vemv:
* Remove `repo` caching
  * See https://github.com/clojure-emacs/cider-nrepl/commit/178556ebfeaf8719e197b7b6e128efe5f85f1a50#diff-f3e9d251e83fc546f51aa7cfc8ae605d64f20c232423c4b4eaf968f227b09ac6
* Cache the uncompressed JDK sources (see `project.clj`, JDK-8 only)
* CI: Use the JDK as part of the cache key
  * Especially important for the `unzipped-jdk-source` stuff.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,13 +64,19 @@ commands:
     description: |
       Run a set of steps with Maven dependencies and Clojure classpath cache
       files cached.
-      This command restores ~/.m2 and .cpcache if they were previously cached,
+      This command restores ~/.m2, .cpcache and such if they were previously cached,
       then runs the provided steps, and finally saves the cache.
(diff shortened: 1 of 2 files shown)


vemv:
PR ready, sorry for the excess commits

Looks like the `unzipped-jdk-source` stuff wasn't the culprit for the few jobs that take 2+ minutes.

Will investigate, as it can plausibly be an actual issue in the codebase.

vemv:
Hmm, briefly checking out the related source code, assuming that the slowness comes from `@cache-initializer`, the related code invokes `immutable-source-file?` which @plexus found out to be slow.

Let's see how that one PR ends up :)

bbatsov (approved):

PR #132, opened 2021-09-30
0.7.2

vemv:
I'd like the latest batch of bugfixes to make it to cider-nrepl, which we intend to release as an alpha today / soon.

LGTY? @bbatsov


bbatsov:
Agreed.

Issue #133, opened 2021-10-01
Regression in eldoc 


vemv:
https://app.circleci.com/pipelines/github/clojure-emacs/cider-nrepl/448/workflows/ab0029aa-c7c9-45d2-9071-d5692ea11aad/jobs/3661

i.e. querying eldoc for `non.existing/+` will show the doc for `clojure.core/+`.

This is caused by https://github.com/clojure-emacs/orchard/commit/7d7f752e5d1abd1cdfbbcd164a7a5ca955d7789d

It's not a critical bug, anyway @plexus could you please take a look?

plexus:
How do you suggest addressing that, short of reverting that change?

vemv:
Perhaps we can make the change opt-out via an option and make `cider.nrepl.middleware.info` opt out?

bbatsov:
I don't think we need to overcomplicate this. There are already a ton of config options that nobody knows/uses. 

> i.e. querying eldoc for non.existing/+ will show the doc for clojure.core/+.

I though this was the intention of the change, so perhaps we should just update the tests in `cider-nrepl` to reflect this.

vemv:
I don't think showing a wrong eldoc for a different ns is correct behavior.

i.e. if I'm editing a given ns, it's fair enough to assume that an unqualified `+` refers to `clojure.core/+`.

But if querying a different ns, `other/+` it's **definitionally** something other than `clojure.core/+`. It cannot equal to `clojure.core/+`. (If it was identical, why would it be defined at all?)

bbatsov:
But you can't know this before the `other` ns was actually evaluated, that's why I don't see much of a problem with this. You might get a false positive in an unevaluated ns, but you'd be fine otherwise. Am I missing something?

vemv:
One cannot necessarily know if the ns in question is evaluated... with tools.namespace is less obvious than it sounds (e.g. it can unload namespaces, fail, leave things uneval'd)

I'll check out again the logic/integration in question. Maybe we can easily distinguish these two cases?

* a ns name was not passed (infer OK)
* a ns name was passed (infer NOK)

plexus:
I think you're right, we can limit this fallback behavior to unqualified symbols.

bbatsov:
Makes sense to me as well.

PR #134, opened 2021-10-02
`info`: don't use a fallback for ns-qualified symbols

vemv:
* `info`: don't use a fallback for ns-qualified symbols
  * Closes https://github.com/clojure-emacs/orchard/issues/133
  * ℹ️ cider-nrepl will still need to get a couple tests adapted (and probably a couple more _added_)
* Add `make install` command
  * Also includes our usual refinements (`-user` etc).
* Unflake a test

### DIFF
--- a/src/orchard/info.clj
+++ b/src/orchard/info.clj
@@ -63,7 +63,7 @@
 
 (defn clj-meta
   {:added "0.5"}
-  [{:keys [dialect ns sym computed-ns unqualified-sym] :as opts}]
+  [{:keys [dialect ns sym computed-ns unqualified-sym]}]
   {:pre [(= dialect :clj)]}
   (let [ns (or ns computed-ns)
         ns (or (when (some-> ns find-ns)
--- a/src/orchard/info.clj
+++ b/src/orchard/info.clj
@@ -7,9 +7,9 @@
    [orchard.cljs.meta :as cljs-meta]
    [orchard.java :as java]
    [orchard.java.classpath :as cp]
+   [orchard.java.resource :as resource]
    [orchard.meta :as m]
-   [orchard.misc :as misc]
-   [orchard.java.resource :as resource]))
+   [orchard.misc :as misc]))
 
 (defn qualify-sym
   "Qualify a symbol, if any in `sym`, with `ns`.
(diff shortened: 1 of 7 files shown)


src/orchard/info.clj
```diff
@@ -65,32 +65,29 @@
...
-      (recur (assoc opts :ns 'clojure.core))
-      (or
-       ;; it's a special (special-symbol?)
-       (m/special-sym-meta sym)
-       ;; it's a var
-       (some-> ns (m/resolve-var sym) (m/var-meta))
-       ;; it's a Java constructor/static member symbol
-       (some-> ns (java/resolve-symbol sym))
-       ;; it's an unqualified sym maybe referred
-       (some-> ns (m/resolve-var unqualified-sym) (m/var-meta))
```
vemv, line 81: Removed these two lines, I don't think they make sense. If one is explicitly querying `:sym foo/+`, then trying to resolve `+` without foo is not accurate.

The comment says `maybe referred`.  But `(some-> ns (m/resolve-var sym) (m/var-meta))` already handles refers, that's part of what `resolve/var`/`ns-resolve` does

test/orchard/info_test.clj
```diff
@@ -368,7 +369,7 @@
                     (map #(select-keys % [:ns :name :arglists :macro :file]))))))
 
       (testing "- :clj"
-        (is (= (take 2 (repeat expected))
+        (is (= [{}, expected]
```
vemv, line 372: Changed this expectation, `orchard.test-ns/my-add` does not exist so the right result for it is `{}`

test/orchard/info_test.clj
```diff
@@ -467,21 +471,49 @@
...
-                        [:added :ns :name :file]))))
-  (testing "clojure.core syms can still be resolved"
-    (is (= '{:added "1.0"
-             :ns clojure.core
-             :name merge
-             :file "clojure/core.clj"}
-           (select-keys (info/info* {:ns 'gibberish :sym 'merge})
-                        [:added :ns :name :file])))))
+  (let [current-ns (-> ::_ namespace symbol)]
+    (are [input expected] (= expected
```
bbatsov, line 482: I don't like `are` as I think it reduces the readability of tests. For me it's better to have more `testing` blocks, as they make it easier to follow what a group of assertions tests exactly.

test/orchard/java/resource_test.clj
```diff
@@ -8,6 +8,8 @@
 
 (deftest project-resources-test
   (testing "get the correct resources for the orchard project"
-    (let [resources (resource/project-resources)]
+    (let [resources (->> (resource/project-resources)
```
bbatsov, line 11: What was wrong with this test before? There were more resources returned or something?

vemv: >  There were more resources returned or something?

Yes especially in local envs were there can be more resources for arbitrary reasons.

test/orchard/info_test.clj
vemv: Perhaps we can agree to disagree? I didn't use an `are` of the crazy flavor here. It's a simple mapping of inputs->outputs.

Normally to reach this level of exhaustiveness with manually coded tests, one would have to take 1 to 2 'screenfuls' worth of code. That's generally not too readable - it's so wordy that it's hard to compare the test cases between them.

Normally people give up before even attempting this level of exhaustiveness, so `are` can invite us to do the right thing (focus on the data, compare the tests cases between them, cheaply add more cases if missing)

bbatsov: I can live with this, I simply don't like it. Too many years of writing RSpecs I guess. :-) There's also the practical problem that CIDER's test runner doesn't work well with `are`, but that's not a big deal.

test/orchard/java/resource_test.clj
bbatsov: Got it.

vemv:
0.7.3 released with these changes and consumed in cider-nrepl https://github.com/clojure-emacs/cider-nrepl/pull/715

plexus:
Thanks for picking this up so quickly!

vemv:
🍻 Turned out that https://github.com/clojure-emacs/orchard/commit/7d7f752e5d1abd1cdfbbcd164a7a5ca955d7789d wasn't wrong per se but it helped surface various edge cases (covered with tests here and downstream)

PR #138, opened 2021-11-05
Remove dynapath - WIP

vemv:

- [ ] All tests are passing

Thanks!


src/orchard/java.clj
```diff
@@ -141,16 +116,15 @@
   (if (>= misc/java-api-version 9)
     (do (require '[orchard.java.parser :as src])
         (resolve 'src/source-info))
-    (if jdk-tools
-      (do (require '[orchard.java.legacy-parser :as src])
-          (resolve 'src/source-info))
-      (constantly nil))))
+    (do
+      (assert jdk-tools)
```
vemv, line 120: Probably the `assert` is dangerous here, we cannot be sure of whether Orchard is being used with a plugin that will  place the tools on the classpath.

Relatedly, in cider.el it can be hard to conditionally inject the tools plugin depending on the detected JDK. 

For these reasons it seems good to move the tools logic to enrich-classpath which would do the JDK detection.

src/orchard/java/classpath.clj
```diff
@@ -70,12 +67,7 @@
   "Adds the URL to the classpath and returns it if successful, or nil otherwise,
   ensuring that a modifiable classloader is available."
   [^URL url]
-  (let [loader (or (modifiable-classloader)
```
vemv, line 73: Mark as deprecated, maybe even fail-fast

src/orchard/java/classpath.clj
```diff
@@ -32,11 +32,8 @@
    (classloaders (context-classloader))))
 
 (defn modifiable-classloader
-  "Returns the highest classloader in the hierarchy that satisfies
-  `dynapath.util/addable-classpath?`, or nil if none do."
   ([^ClassLoader loader]
-   (last (filter dp/addable-classpath?
-                 (classloaders loader))))
+   loader)
```
vemv, line 36: Mark as deprecated, maybe even fail-fast

project.clj
```diff
@@ -104,19 +103,21 @@
                                      [org.clojure/clojure "1.11.0-master-SNAPSHOT" :classifier "sources"]]}
 
              :test {:dependencies [[org.clojure/java.classpath "1.0.0"]]
-                    :resource-paths ["test-resources"
-                                     "not-a.jar"
-                                     "does-not-exist.jar"]
+                    :resource-paths ~(cond-> ["test-resources"
+                                              "not-a.jar"
+                                              "does-not-exist.jar"]
+                                       jdk8? (conj (unzipped-jdk-source))
```
vemv, line 109: Instead of this I should use enrich-classpath for realism

src/orchard/java.clj
bbatsov: Yeah, that makes perfect sense, but I guess it will need at least clojure cli support first.

bbatsov: Btw, do we need an `assert` here - it might a bit frustrating for users that Orchard will just blow up there. Perhaps the conditional approach that does nothing if `jdk-tools` is missing is better?

vemv: Yeah that's what I meant w/ `Probably the assert is dangerous here`. Probably when I wrote this code a few weeks ago I meant the assert as a "fix this before merging it" thing i.e. a glorified TODO 😄

btw, yesterday I progressed a little https://github.com/clojure-emacs/enrich-classpath/commit/4cb44de5502494b9806aa570cc410ed87ffbcf6b

bbatsov:
We should cut 0.8 after this is merged.

src/orchard/java/classpath.clj
vemv: probably fail-fast is risky, similarly to `assert`. An end user could be used the 'deprecated' path unintentionally.

vemv:
> We should cut 0.8 after this is merged.

final round for enrich-classpath coming soon! :)

project.clj
vemv, line 109: "./unzipped-jdk-source/"  should be something like `~/.cider/orchard/<hash of jdk>/unzipped-jdk-source`

project.clj
```diff
@@ -104,19 +103,21 @@
...
+                                       ;; must go last
+                                       ;; must be unconditional - Orchard must work when src.zip is present;
+                                       ;; for w/e reason:
+                                       true (conj jdk-sources))
+                    :plugins ~(if jdk8?
+                                '[[lein-jdk-tools "0.1.1"]]
```
vemv, line 115: rm this plugin, as enrich-classpath now does this

project.clj
bbatsov: Won't <version-of-the-jdk> be a bit more readable than a hash? I

vemv: > Won't be a bit more readable than a hash?

Finally I'm doing `(->> "java.version" System/getProperty (re-seq #"\d+") (string/join))` as of my wip

vemv:
CI Matrix is passing with Enrich and no dynapath ❤️

Will wrap this up over the week, very few details remaining.

vemv:
Will reopen from scratch

Issue #122, opened 2021-08-27
Create an integration test with enrich-classpath


vemv:
It would be nice to ensure in tests that https://github.com/clojure-emacs/enrich-classpath/ allows `orchard.java/member-info` to return info that it otherwise wouldn't.

I'd do it by creating a separate profile / CI step just for this integration test. Otherwise it's not great to have a non-essential dependency present for the whole suite.

Issue #120, opened 2021-08-11
Remove dynapath


bbatsov:
Mostly a reminder for me and @vemv. Now that enrich-classpath is ready for action with can proceed with our plan and remove the dynapath-related functionality completely from Orchard.

vemv:
👍!

I have to give another pass to enrich-classpath though. All three open issues are at hand, simply I have to find a good chance to execute

Issue #105, opened 2021-02-02
Orchard error on thrown error: java.lang.ClassNotFoundException: com.sun.tools.javac.util.List


danielneal:
## Expected behavior
`(throw (ex-info "Help!" {})` should be handled by cider and show *cider-error*

## Actual behavior
Execution error (ExceptionInfo) at user/eval9510 (REPL:1265).
Helo!
ERROR: Unhandled REPL handler exception processing message {:nrepl.middleware.print/stream? 1, :nrepl.middleware.print/print cider.nrepl.pprint/pprint, :nrepl.middleware.print/quota 1048576, :nrepl.middleware.print/buffer-size 4096, :nrepl.middleware.print/options {:right-margin 70}, :op stacktrace, :session a739d2ab-eec5-42ae-8e35-01b1ea428624, :id 234}
java.lang.ClassNotFoundException: com.sun.tools.javac.util.List
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at clojure.lang.DynamicClassLoader.findClass(DynamicClassLoader.java:69)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at clojure.lang.DynamicClassLoader.loadClass(DynamicClassLoader.java:77)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at clojure.lang.RT.classForName(RT.java:2211)
	at clojure.lang.RT.classForName(RT.java:2220)
	at cider.nrepl.inlined_deps.orchard.v0v6v4.orchard.java.legacy_parser$parse_java.invokeStatic(legacy_parser.clj:88)
	at cider.nrepl.inlined_deps.orchard.v0v6v4.orchard.java.legacy_parser$parse_java.invoke(legacy_parser.clj:66)

## Steps to reproduce the problem

1. Set up a new tools.deps project. 
2. Throw an error

### Clojure version
1.10.1

### Java version
java -version  
openjdk version "1.8.0_275"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_275-b01)

### Operating system
Mac OS X 10.15.7

### Emacs version
GNU Emacs 27.1 (build 1, x86_64-apple-darwin19.6.0, Carbon Version 162 AppKit 1894.6) of 2021-02-02

## Notes
[...]

Issue #50, opened 2019-04-24
Ensure CI image includes JDK sources and tools 


jeffvalk:
To run all tests, the CI Docker image needs to include:

1. `src.zip` for all JDKs; and
2. `tools.jar` for JDK8.

Both of these should be located either in the JDK root, or in the `lib` directory under this.

shen-tian:
There's a growing list of reason to look beyond the CircleCI convenience images for Clojure. I'd be up for having a go at that from mid may or so, and that should address this. But anyone else who can get to this before then, please do :)

bbatsov:
Btw, an user just reported on Slack that with the current Orchard snapshot he's getting:

```
Syntax error compiling at (cider/nrepl/inlined_deps/orchard/v0v5v0_beta2/orchard/java/legacy_parser.clj:1:1).
...
Caused by: java.lang.ClassNotFoundException: com.sun.javadoc.ClassDoc
    at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
    at clojure.lang.DynamicClassLoader.findClass(DynamicClassLoader.java:69)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    at clojure.lang.DynamicClassLoader.loadClass(DynamicClassLoader.java:77)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    at java.lang.Class.forName0(Native Method)
    at java.lang.Class.forName(Class.java:348)
    at clojure.lang.RT.classForName(RT.java:2211)
    at clojure.lang.RT.classForNameNonLoading(RT.java:2224)
    at cider.nrepl.inlined_deps.orchard.v0v5v0_beta2.orchard.java.legac...
```

I assume the reason is the same and I'll relay the solution, but I think we should try to fail more gracefully in such cases.

jeffvalk:
If the user could confirm or refute the hypothesis that this is caused by a bad file on the classpath, that would be helpful.

bbatsov:
Btw, this error is also on the cider-nrepl builds now https://circleci.com/gh/clojure-emacs/cider-nrepl/151

bbatsov:
And this has also cascaded to cljdoc https://circleci.com/gh/cljdoc/builder/4505 :-)

jeffvalk:
Presumably the same image is used for the CI container in all of those cases. Seeing what's in that sounds like the place to start.

bbatsov:
[...]

PR #140, opened 2021-12-13
Remove `dynapath`, accomodate `enrich-classpath`

vemv:
Fixes https://github.com/clojure-emacs/orchard/issues/122
Fixes https://github.com/clojure-emacs/orchard/issues/120
Fixes https://github.com/clojure-emacs/orchard/issues/105
Fixes https://github.com/clojure-emacs/orchard/issues/103
Fixes https://github.com/clojure-emacs/orchard/issues/50


src/orchard/java.clj
```diff
@@ -44,58 +42,27 @@
...
-                     (some-> (io/resource path)
-                             ^JarURLConnection (. openConnection)
-                             (. getJarFileURL)))]
-      (or (base-url "java.base/java/lang/Object.java") ; JDK9+
-          (base-url "java/lang/Object.java")           ; JDK8-
-          (jdk-find "src.zip")))))
+  [_]
+  nil)
+
+(def ^:deprecated jdk-sources nil)
```
vemv, line 53: From what I pondered on previous iterations: ^:deprecated + no-op impls are the least breaking choice

(as opposed to an `(assert false)`)

src/orchard/java/classpath.clj
```diff
@@ -56,26 +50,26 @@
        (.split (System/getProperty "java.class.path")
                (System/getProperty "path.separator"))))
 
+(defn classpath-urls [classloader]
```
vemv, line 53: This is (essentially) what `dp/classpath-urls` was doing

test-newer-jdks/orchard/java/parser_test.clj
```diff
@@ -1,66 +1,50 @@
 (ns orchard.java.parser-test
   (:require
+   [clojure.test :refer [deftest is testing]]
    [orchard.java.parser :as parser]
-   [clojure.test :refer [deftest is testing]]))
+   [orchard.test.util :as util])
+  (:import
+   (orchard.java DummyClass)))
 
-(defn compile-class-from-source
```
vemv, line 6: Now compilation for this test case happens via Lein :java-source-paths, it reflects typical end-user usage and also exercises what enrich-classpath does / is able to do (it's not able to modify classpaths at runtime, and we don't want that either!)

bbatsov (approved):
Looks good  to me! Great work!

PR #141, opened 2022-01-02
fn-deps for lambdas

### LINKED ISSUE #51: Implement find-usages
Cognitect recently released a [new version of REBL](https://www.patreon.com/posts/rebl-0-9-157-var-26478905) that has a find usages feature. @SevereOverfl0w asked Rich about the implementation and Rich was kind enough to share the crux of it:

```
(defn fdeps [val]
    (set (some->> val class .getDeclaredFields
                  (keep (fn [^java.lang.reflect.Field f]
                          (or (and (identical? clojure.lang.Var (.getType f))
                                   (java.lang.reflect.Modifier/isPublic (.getModifiers f))
                                   (java.lang.reflect.Modifier/isStatic (.getModifiers f))
                                   (-> f .getName (.startsWith "const__"))                                  
                                   (.get f val))
                              nil))))))
```

Some notes from Rich:

> It reflects against the fn classes themselves, every closure put fields in for vars,
> that’s what makes vars fast.
> So these are ‘runtime’ deps, not source deps. You won’t see inlined fns or macros.
> fdeps works on the fn value, not the var.

So people have already been building cool things with that idea https://gist.github.com/vlaaad/0350b61e127e82165d195b490999ec0a

So, what we need to do now is add this to Orchard and then wrap it in CIDER nREPL. That should be pretty easy and have a huge impact for the end users.

Cyrik:
Fixes #51 

Uses Clojures internal class cache to also check anonymous function dependencies.

### DIFF
--- a/src/orchard/xref.clj
+++ b/src/orchard/xref.clj
@@ -105,7 +105,7 @@
 (defn fn-deps [s]
   (when-let [^clojure.lang.AFn v (as-val s)]
     (let [f-class-name (-> v .getClass .getName)
-          ^java.lang.reflect.Field field (->> clojure.lang.DynamicClassLoader .getDeclaredFields second)
+          field (->> clojure.lang.DynamicClassLoader .getDeclaredFields second)
           classes (into {} (.get field clojure.lang.DynamicClassLoader))
           filtered-classes (->> classes
                                 (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
--- a/src/orchard/xref.clj
+++ b/src/orchard/xref.clj
@@ -35,7 +35,7 @@
                                    (.get f (fn-name v)))
                               nil)))))))
 
-(def ^:private classCache
+(def ^:private class-cache
   "Reference to Clojures class cache.
    This holds of classes compiled by the Clojure compiler,
    one class per function and one per repl eval.
@@ -62,7 +62,7 @@
       (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
                       (map (fn [[_k v]] (.get ^java.lang.ref.Reference v)))
                       (mapcat fn-deps-class))
-            classCache))))
+            class-cache))))
 
 (defn- fn->sym
   "Convert a function value `f` to symbol."
@@ -106,7 +106,7 @@
   (let [f-class-name "orchard.xref" #_(-> orchard.xref/fn-deps .getClass .getName)
         classes (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
                                 (map (fn [[_k v]] (.get ^java.lang.ref.Reference v))))
-                      classCache)]
+                      class-cache)]
     classes)
 
   (let [memKiller (java.util.ArrayList.)]
--- a/src/orchard/xref.clj
+++ b/src/orchard/xref.clj
@@ -4,6 +4,7 @@
   {:added "0.5"}
   (:require
    [clojure.repl :as repl]
+   [clojure.set :as set]
    [clojure.string :as str]
    [orchard.query :as q]))
 
@@ -21,7 +22,7 @@
 (defn fn-deps-class
   "Returns a set with all the functions invoked by `v`.
   `v` can be a function class or a symbol."
-  {:added "0.8"}
+  {:added "0.9"}
   [v]
   (let [^java.lang.Class v (if (class? v)
                              v
@@ -64,6 +65,23 @@
                       (mapcat fn-deps-class))
             class-cache))))
 
+(defn fn-transitive-deps
+  "Returns a set with all the functions invoked inside `v` or inside those functions.
+  `v` can be a function value, a var or a symbol."
+  {:added "0.9"}
+  [v]
+  (let [deps (fn-deps v)]
+    (loop [checked #{}
+           to-check deps
+           deps deps]
+      (cond
+        (empty? to-check) deps
+        :else (let [[current & remaining] to-check
+                    new-deps (fn-deps current)]
+                (recur (conj checked current)
+                       (concat remaining (filter #(contains? deps %) new-deps))
+                       (set/union deps new-deps)))))))
+
 (defn- fn->sym
   "Convert a function value `f` to symbol."
   [f]
@@ -88,6 +106,7 @@
     (map first (filter (fn [[_k v]] (contains? v var)) deps-map))))
 
 (comment
+  ;; this can be used to blow up memory, which will clear the class cache of old references
   (defn oom []
     (try (let [memKiller (java.util.ArrayList.)]
            (loop [free 10000000]
@@ -103,18 +122,14 @@
   (fn-deps #'orchard.xref/fn-deps)
   (fn-refs #'orchard.xref/fn->sym)
 
+  ;; returns all classes in this ns, even repl eval'd
   (let [f-class-name "orchard.xref" #_(-> orchard.xref/fn-deps .getClass .getName)
         classes (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
                                 (map (fn [[_k v]] (.get ^java.lang.ref.Reference v))))
                       class-cache)]
     classes)
 
-  (let [memKiller (java.util.ArrayList.)]
-    (loop [free (.. Runtime (getRuntime) (freeMemory))]
-      (.add memKiller (object-array free))
-      (recur (.. Runtime (getRuntime) (freeMemory)))))
   (oom)
-  (Math/min 1 2)
   (def vars (q/vars {:ns-query {:project? true} :private? true}))
 
   (map fn-deps vars))
--- a/src/orchard/xref.clj
+++ b/src/orchard/xref.clj
@@ -7,23 +7,6 @@
    [clojure.string :as str]
    [orchard.query :as q]))
 
-(defn- hunt-down-source
-  [fn-sym]
-  (let [{:keys [source file line]} (-> fn-sym
-                                       resolve
-                                       meta)]
-    (try (or source
-             (and file (read-string {:read-cond :allow}
-                                    (or
-                                     (clojure.repl/source-fn fn-sym)
-                                     (->> file
-                                          slurp
-                                          clojure.string/split-lines
-                                          (drop (dec line))
-                                          (clojure.string/join "\n"))
-                                     "nil"))))
-         (catch Exception _)))) ;; explodes on namespaced keywords
-
 (defn- as-val
   "Convert `thing` to a function value."
   [thing]
@@ -32,47 +15,8 @@
     (symbol? thing) (var-get (find-var thing))
     (fn? thing) thing))
 
-(defn- f->sym [^clojure.lang.AFn f]
-  (-> f .getClass .getName repl/demunge symbol))
-
-(defn- fn-source [f]
-  (hunt-down-source (f->sym f)))
-
 (defn- fn-name [^java.lang.Class f]
   (-> f .getName repl/demunge symbol))
-
-(defonce classbytes (atom {}))
-
-(defn- recompile [ns-sym form]
-  (push-thread-bindings
-   {clojure.lang.Compiler/LOADER
-    (proxy [clojure.lang.DynamicClassLoader] [@clojure.lang.Compiler/LOADER]
-      (defineClass
-        ([name bytes src]
-         (swap! classbytes assoc name bytes)
-         (proxy-super defineClass name bytes src))))})
-  (try
-    (let [line @clojure.lang.Compiler/LINE
-          column @clojure.lang.Compiler/COLUMN
-          line (if-let [line (:line (meta form))]
-                 line
-                 line)
-          column (if-let [column (:column (meta form))]
-                   column
-                   column)]
-      (push-thread-bindings {clojure.lang.Compiler/LINE line
-                             clojure.lang.Compiler/COLUMN column})
-      (try
-        (let [form (macroexpand form)]
-          (when (and (coll? form) (= 'clojure.core/fn (first (nth form 2 nil))))
-            (binding [*ns* (create-ns ns-sym)]
-              (clojure.lang.Compiler/analyze
-               clojure.lang.Compiler$C/EVAL
-               (nth form 2)))))
-        (finally
-          (pop-thread-bindings))))
-    (finally
-      (pop-thread-bindings))))
 
 (defn- fn-deps-class
   [v]
@@ -88,25 +32,17 @@
                                    (.get f (fn-name v)))
                               nil)))))))
 
-(defn fn-deps-compiling
+(defn fn-deps 
   "Returns a set with all the functions invoked by `val`.
   `val` can be a function value, a var or a symbol."
   {:added "0.5"}
-  [f]
-  (reset! classbytes {})
-  (let [f (as-val f)]
-    (when-let [source (and (fn? f) (fn-source f))]
-      (let [sym (f->sym f)
-            b (do (recompile (-> sym namespace symbol) source) @classbytes)
-            class-names (map first b)
-            deps (set (mapcat #(-> % symbol fn-deps-class) class-names))]
-        deps))))
-
-(defn fn-deps [s]
+  [s]
   (when-let [^clojure.lang.AFn v (as-val s)]
     (let [f-class-name (-> v .getClass .getName)
-          field (->> clojure.lang.DynamicClassLoader .getDeclaredFields second)
-          classes (into {} (.get field clojure.lang.DynamicClassLoader))
+          ;; breaks when called with
+          ;; (.getDeclaredField clojure.lang.DynamicClassLoader "classCache")
+          classCache (->> clojure.lang.DynamicClassLoader .getDeclaredFields second)
+          classes (into {} (.get classCache clojure.lang.DynamicClassLoader))
           filtered-classes (->> classes
                                 (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
                                 (map (fn [[_k v]] (.get ^java.lang.ref.Reference v))))
@@ -138,10 +74,9 @@
 
 (comment
   (nth (macroexpand (read-string "(def archive? (clojure.core/fn ([f] (file-ext? f .jar .zip))))")) 2)
-  (hunt-down-source 'orchard.util.os-test/cache-dir-windows-test)
   (fn-deps #'fn-refs)
   (fn-deps #'orchard.xref/fn->sym)
   (fn-refs #'orchard.xref/fn->sym)
-  (supers (type @clojure.lang.Compiler/LOADER))
+  (into {} (.get (->> clojure.lang.DynamicClassLoader .getDeclaredFields second) clojure.lang.DynamicClassLoader))
   (def vars (q/vars {:ns-query {:project? true} :private? true}))
   (map fn-deps vars))
--- a/src/orchard/xref.clj
+++ b/src/orchard/xref.clj
@@ -3,6 +3,7 @@
   references."
   {:added "0.5"}
   (:require
+   [clojure.string]
    [clojure.repl :as repl]
    [orchard.query :as q]))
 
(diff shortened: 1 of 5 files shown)


bbatsov:
Small note - we can't really depend on `tools.reader`, as this might create conflicts with user applications. Orchard has to have no runtime deps.

Cyrik:
> Small note - we can't really depend on tools.reader, as this might create conflicts with user applications.

ah thanks, I completely forgot I was still depending on that. I'll rewrite that part to use clojure/read.

Do you have any thoughts on the idea of using the compiler to find the dependencies?

bbatsov:
The idea itself seems reasonable to me.

Cyrik:
I've got a new solution that doesn't need to recompile the code and is a lot faster.
The problem with the new solution is that old classes stay around in clojures cache for some time. I haven't found a way yet to find out which old classes aren't in use anymore.
But the new solution is already a lot better than the old one.
Do you have any ideas on how to fix the final problem?
Also would you want to keep the recompile version as an extra possibility if the actual bytecode is needed for something else?

vemv:
> The problem with the new solution is that old classes stay around in clojures cache for some time. 

Can you explain this as a PR comment (self-review)? That way we can see the code that the problem in question relates to.

src/orchard/xref.clj
```diff
@@ -13,21 +32,87 @@
...
+            b (do (recompile (-> sym namespace symbol) source) @classbytes)
+            class-names (map first b)
+            deps (set (mapcat #(-> % symbol fn-deps-class) class-names))]
+        deps))))
+
+(defn fn-deps [s]
+  (when-let [^clojure.lang.AFn v (as-val s)]
+    (let [f-class-name (-> v .getClass .getName)
+          field (->> clojure.lang.DynamicClassLoader .getDeclaredFields second)
+          classes (into {} (.get field clojure.lang.DynamicClassLoader))
```
Cyrik, line 109: this holds all functions compiled by clojure:
{...
"dev$dummy"
   #object
    [java.lang.ref.SoftReference 0x5081395f
     "java.lang.ref.SoftReference@5081395f"]
...
}
where lambdas are named "dev$dummy$fn_1234"

the filter uses that fact to just keep the function + lambda classes.

The problem is, when you recompile a function, new lambda classes get generated and the old ones stay around.

src/orchard/xref.clj
```diff
@@ -13,21 +32,87 @@
...
+      (let [sym (f->sym f)
+            b (do (recompile (-> sym namespace symbol) source) @classbytes)
+            class-names (map first b)
+            deps (set (mapcat #(-> % symbol fn-deps-class) class-names))]
+        deps))))
+
+(defn fn-deps [s]
+  (when-let [^clojure.lang.AFn v (as-val s)]
+    (let [f-class-name (-> v .getClass .getName)
+          field (->> clojure.lang.DynamicClassLoader .getDeclaredFields second)
```
vemv, line 108: can you please get the field by name? so that we can see what field it is

src/orchard/xref.clj
vemv: >  the old ones stay around

As in, garbage to be collected?

What causes that retention? This is a `defn` (and not e.g. a `def`), so why invoking a defn would retain memory references?

Unless there's global state somewhere, probably your solution is good - it's just minor details like `*1` or other repl state that might retain objects for longer than expected.

src/orchard/xref.clj
Cyrik: that's a problem I'm still having, if I change that line to (.getDeclaredField clojure.lang.DynamicClassLoader "classCache") which is basically the same thing, I get an IllegalAccessException on the next line.
The same thing happens if I type hint field to ^java.lang.reflect.Field

src/orchard/xref.clj
```diff
@@ -13,21 +32,87 @@
...
+(defn- f->sym [^clojure.lang.AFn f]
+  (-> f .getClass .getName repl/demunge symbol))
+
+(defn- fn-source [f]
+  (hunt-down-source (f->sym f)))
+
+(defn- fn-name [^java.lang.Class f]
+  (-> f .getName repl/demunge symbol))
+
+(defonce classbytes (atom {}))
```
vemv, line 44: We shouldn't have global state, sounds like a recipe for memory leaks

src/orchard/xref.clj
```diff
@@ -13,21 +32,87 @@
...
+
+(defn- recompile [ns-sym form]
```
vemv, line 46: Why is "recompilation" needed? This would be a great `;; comment` to add.

src/orchard/xref.clj
```diff
@@ -13,21 +32,87 @@
...
+  (push-thread-bindings
+   {clojure.lang.Compiler/LOADER
+    (proxy [clojure.lang.DynamicClassLoader] [@clojure.lang.Compiler/LOADER]
```
vemv, line 49: Just as a word of caution, hacking with classloaders as a long history of giving us problems, in various tools over the years.

If you have the chance, please avoid these as they can compose badly with other features.

Clojure internals are best kept internal :)

src/orchard/xref.clj
Cyrik: I probably explained it wrong.
The problem is not the memory, the problem is that the Clojure classloader keeps the old classes in its own cache as java.lang.ref.SoftReference and only deletes that if you try to access the reference after it has been cleaned up.
So if I do the following in my code:
```clojure
(defn dummy []
  (map #(println %) (range 10)))
(defn dummy []
  (map #(print %) (range 10)))
;;o r just send both versions to the repl by changing the first
```
classes will now contain "dev$dummy$fn_123" and "dev$dummy$fn_124" where 123 references println and 124 references print.
I haven't found an easy way to get rid of "dev$dummy$fn_123" yet.

src/orchard/xref.clj
vemv: At least `field` could be named more expressively

src/orchard/xref.clj
vemv: Perhaps it would be solved if WeakReference was used instead.

However, as mentioned personally I do not condone classloader hacking :)

src/orchard/xref.clj
Cyrik: You are right, I should have explained it in the code.
The idea is to recompile the code and overwrite the classLoader just during that recompilation so that we can get at the bytecode. Sadly there is no other way to do that since the JVM just throws all of that away after loading.
This "should" not interfere with normal execution since the new classes are never used and the proxied loader is thrown away after use.
But hopefully, the recompile solution won't be needed if I can fix the final problem with the second solution.
We might want to keep the code around though since the bytecode can be used for other introspection at a later point.

src/orchard/xref.clj
Cyrik: This solution does not overwrite the clojure classLoader, so I can't switch the reference type. This just accesses the regular clojure loader through reflection.
I probably shouldn't have kept two separate solutions in the same file, I can see how that is confusing, sorry.

vemv:
Part of the context is https://lukas-domagala.de/blog/clojure-analysis-and-introspection.html which says:

> [clj-kondo] works great but has two new problems:
> * it's kind of slow (40 seconds for that project, including jars)
> * does not work for running processes where you don't have the code (remote repls)

Personally I like the middle ground of using clj-kondo but spawn from the same JVM that hosts a given repl.

That way the classpath is always up to date, and one doesn't need external binaries.

clj-kondo itself uses a persistent cache so I don't believe saying it takes 40s to analyze a project is quite representative of day-to-day UX.

As you, I much favor runtime-powered solutions where they make sense. Rich Hickey's hack is a hack, it might work for him because he authored the Clojure compiler.

The compiler's internals are generally not for us to rely on.

So clj-kondo is a fine solution, just as tools.analyzer is (which is very much runtime-powered, so survives any dynamism).

My take would be that unless this PR becomes substantially more minimalistic, it would be not be worth the hassle.

One can always use his own tooling on a personal stack - which I do all the time! Later one can propose PRs as those solutions become more time-proven.

Cyrik:
Thank you for your comments @vemv
I completely understand your position. Clj-kondo is great, it's just a little slow for my use case, which it was not meant to handle.

I've removed the recompile part of the code and commented on the weird IllegalAccess workaround. It's completely fine if you don't want to merge this. This way whoever stumbles over this will have a clean thing to build upon.

src/orchard/xref.clj
```diff
@@ -13,21 +15,40 @@
...
+(defn fn-deps
+  "Returns a set with all the functions invoked by `val`.
+  `val` can be a function value, a var or a symbol."
+  {:added "0.5"}
+  [s]
+  (when-let [^clojure.lang.AFn v (as-val s)]
+    (let [f-class-name (-> v .getClass .getName)
+          ;; breaks when called with
+          ;; (.getDeclaredField clojure.lang.DynamicClassLoader "classCache")
+          classCache (->> clojure.lang.DynamicClassLoader .getDeclaredFields second)
```
vemv, line 44: Maybe you can `(->> clojure.lang.DynamicClassLoader .getDeclaredFields (filter ...) first)`? Since `second` seems non-deterministic.

Also, the work can be done in a `def` to not repeat the work on each invocation.

vemv, line 44: Why not type-hint this? Seems better than skipping Eastwood.

src/orchard/xref.clj
```diff
@@ -13,21 +15,40 @@
...
+          classes (into {} (.get classCache clojure.lang.DynamicClassLoader))
+          filtered-classes (->> classes
+                                (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
+                                (map (fn [[_k v]] (.get ^java.lang.ref.Reference v))))
+          deps (set (mapcat fn-deps-class filtered-classes))]
```
vemv, line 49: Probably using transducers would improve performance, there can be quite a lot of classes in a given JVM session.

vemv (commented):
The PR looks more minimalistic now so we can proceed, thanks!

src/orchard/xref.clj
vemv: > I haven't found an easy way to get rid of "dev$dummy$fn_123" yet.

Does the PR still have this problem?

It might be acceptable to have that problem, if so at least leave a comment explaining what's going on

bbatsov:
> I've removed the recompile part of the code and commented on the weird IllegalAccess workaround. It's completely fine if you don't want to merge this. This way whoever stumbles over this will have a clean thing to build upon.

Well, our built-in find-usages is not going away, so I'm in favor of improving it. I gets @vemv's position but I think it's fine to keep improving what we have given the any alternative will require deps and extra work that no one has been willing to do for years. :-) 

I also think that Clojure's compiler API has been super stable, so the concerns about whether it's a good idea to use its API are probably overdone. Let's drive this PR to the finish line.

src/orchard/xref.clj
Cyrik: If I type-hint this I get an IllegalAccess Exception, I'm guessing it's the same problem as the direct access.
I'll have a look at the bytecode produced by both versions to hopefully find out why one works and the other doesn't, but it's a lower priority than the other things.

src/orchard/xref.clj
Cyrik: Yes, the PR still has the problem. I'll play around with it some more to see if I can find a solution. If I can't find one I'll add a comment describing the issue and open a new ticket once the PR is merged.

src/orchard/xref.clj
```diff
@@ -13,21 +15,47 @@
...
                                    (java.lang.reflect.Modifier/isStatic (.getModifiers f))
                                    (-> f .getName (.startsWith "const__"))
-                                   (.get f val))
+                                   (.get f (fn-name v)))
                               nil)))))))
 
+(def ^java.lang.reflect.Field classCache
+  (let [classCache* (.getDeclaredField clojure.lang.DynamicClassLoader "classCache")]
+    (.setAccessible classCache* true)
+    (.get classCache* clojure.lang.DynamicClassLoader)))
```
Cyrik, line 41: I've managed to fix the IllegalAccess issue. This was something I tried before, not sure why it's working now, but I'm happy.

src/orchard/xref.clj
```diff
@@ -13,21 +15,47 @@
...
                   (keep (fn [^java.lang.reflect.Field f]
                           (or (and (identical? clojure.lang.Var (.getType f))
                                    (java.lang.reflect.Modifier/isPublic (.getModifiers f))
                                    (java.lang.reflect.Modifier/isStatic (.getModifiers f))
                                    (-> f .getName (.startsWith "const__"))
-                                   (.get f val))
+                                   (.get f (fn-name v)))
                               nil)))))))
 
+(def ^java.lang.reflect.Field classCache
```
Cyrik, line 38: the type-hint is wrong, I'll fix that in the next commit

Cyrik:
I'm still trying to fix the "old Lamdba" reference problem but otherwise this is done.
There is a hacky fix where you force an OOM Exception on purpose, but it doesn't always work and feels very dirty. Hopefully I'll find another way, or I'll leave it like this.

vemv:
Kudos for the iteration!

There's `System/gc`, have you tried that? It might make sense to invoke it whenever the user invokes the functionality (as opposed to on a schedule or something similarly invasive).

WIth WeakReference (vs. SoftReference), GCs are even more effective, but I realise we cannot fork Clojure.

Cyrik:
gc does not help, it only clears weakReferences. Softreferences stay around until the memory is needed so badly that an OOM would be triggered.
I've done a few hours of research at this point and could not find any consistent way to clear the old softReferences from the cache. The oom function inside the comment does work sometimes, but it's very hacky and does expand the jvm used memory to the max, so that is not ideal either.
The only other solution I found is [JVM TI](https://docs.oracle.com/javase/1.5.0/docs/guide/jvmti/jvmti.html#IterateOverReachableObjects) but using that from clojure would be very annoying and it's not part of every jvm.

TLDR: I think we are 95% of the way to the full solution and the final 5% are either a lot of work or have big drawbacks.

src/orchard/xref.clj
bbatsov: Just add some comments explaining this for posterity's sake.

bbatsov:
Might also be a good idea to add a section in the README explaining some of the limitations for the fn-deps/refs. I've been meaning to document all the features of Orchard in the README, but I never got to doing so. We have to start somewhere and that's as good of a place as any.

vemv (commented):
Alright, looking good thanks!

Might give it one last look later today.

Could you update the changelog and PR description in the meantime?

Cyrik:
Done.

I also wrote a follow-up [post](https://lukas-domagala.de/blog/clojure-compiler-class-cache.html) on why the class cache is hard to clear and linked that.

src/orchard/xref.clj
```diff
@@ -13,21 +15,55 @@
...
+  (let [classCache* (.getDeclaredField clojure.lang.DynamicClassLoader "classCache")]
+    (.setAccessible classCache* true)
+    (.get classCache* clojure.lang.DynamicClassLoader)))
+
+(defn fn-deps
+  "Returns a set with all the functions invoked by `v` and its lambdas.
+  `v` can be a function value, a var or a symbol.
+   If a function was defined multiple times, old lambda deps will
+   be returned.
+   This does not return functions marked with meta :inline like +
```
vemv, line 54:
```suggestion
   This does not return functions marked with meta :inline like `+`
```

src/orchard/xref.clj
```diff
@@ -13,21 +15,55 @@
...
+   since they are already compiled away at this point."
+  {:added "0.5"}
+  [v]
+  (when-let [^clojure.lang.AFn v (as-val v)]
+    (let [f-class-name (-> v .getClass .getName)]
+      ;; this uses the implementation detail that the clojure compiler always
+      ;; prefixes names of lambdas with the name of its surrounding function class
+      (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
+                      (map (fn [[_k v]] (.get ^java.lang.ref.Reference v)))
```
vemv, line 63: Better if this `v` doesn't shadow the surrounding `v`

src/orchard/xref.clj
```diff
@@ -13,21 +15,55 @@
...
+   one class per function and one per repl eval.
+   This field is package private, so it has to be set to
+   accessible otherwise an IllegalAccess exception would
+   be thrown."
+  (let [classCache* (.getDeclaredField clojure.lang.DynamicClassLoader "classCache")]
+    (.setAccessible classCache* true)
+    (.get classCache* clojure.lang.DynamicClassLoader)))
+
+(defn fn-deps
+  "Returns a set with all the functions invoked by `v` and its lambdas.
```
vemv, line 50: Please use a wording clearer than "its lambdas"

vemv (commented):
Almost there 👍 ty!

Cyrik:
If you think this is relevant to orchard I'll add a `fn-transitive-deps` function to this PR. It would just be a recursive walk over the dependencies, so it might also be ok for users to write that themselfs.

vemv (approved):

test/orchard/xref_test.clj
```diff
@@ -38,4 +41,6 @@
     (is (contains? (into #{} (xref/fn-refs #'map)) #'orchard.xref-test/dummy-fn)))
   (testing "with a symbol"
     (is (= (xref/fn-refs 'orchard.xref-test/dummy-fn) '()))
-    (is (contains? (into #{} (xref/fn-refs #'map)) #'orchard.xref-test/dummy-fn))))
+    (is (contains? (into #{} (xref/fn-refs #'map)) #'orchard.xref-test/dummy-fn)))
+  (testing "with a lambda"
```
bbatsov, line 45: Not sure if this test has any value, given that `times` now appears in all the other tests.

src/orchard/xref.clj
```diff
@@ -45,8 +81,40 @@
...
+
+  (let [memKiller (java.util.ArrayList.)]
+    (loop [free (.. Runtime (getRuntime) (freeMemory))]
+      (.add memKiller (object-array free))
+      (recur (.. Runtime (getRuntime) (freeMemory)))))
+  (oom)
+  (Math/min 1 2)
+  (def vars (q/vars {:ns-query {:project? true} :private? true}))
+
+  (map fn-deps vars))
```
bbatsov, line 120: The final newline seems to be missing.

src/orchard/xref.clj
```diff
@@ -45,8 +81,40 @@
...
   {:added "0.5"}
-  [var]
-  (let [var (as-var var)
+  [v]
+  (let [var (as-var v)
         all-vars (q/vars {:ns-query {:project? true} :private? true})
         deps-map (zipmap all-vars (map fn-deps all-vars))]
     (map first (filter (fn [[_k v]] (contains? v var)) deps-map))))
+
+(comment
```
bbatsov, line 90: Are those some functions you used for test purposes? Might be good idea to add a comment explaining this.

bbatsov:
Looks good to me overall. I even read the blog post - well done! I added a few small remarks and we are good to go. I guess we'll squash this to a single comment in the final version.

Cyrik:
I've added fn-transitive-deps, if you don't want it in here I'll throw it out again or put it in another PR.
Also added one more caveat to the readme. Didn't know this didn't work on AoT code, just found that out because of the transitive test.

test/orchard/xref_test.clj
```diff
@@ -38,4 +44,12 @@
...
+    (is (contains? (into #{} (xref/fn-refs #'times)) #'orchard.xref-test/dummy-fn))))
+
+(deftest fn-transitive-deps-test
+  (testing "basics"
+    (is (= (xref/fn-transitive-deps dummy-fn)
+           #{#'clojure.core/even? #'clojure.core/filter #'clojure.core/map
+             #'clojure.core/range #'orchard.xref-test/times #'orchard.xref-test/times*}))))
```
vemv, line 55:
```suggestion
(deftest fn-transitive-deps-test
  (testing "basics"
    (let [expected #{#'clojure.core/even? #'clojure.core/filter #'clojure.core/map
                     #'clojure.core/range #'orchard.xref-test/times #'orchard.xref-test/times*}]
      (is (contains? expected #'orchard.xref-test/times*)
          "Specifically includes `#'times*`, which is a transitive dep of `#'dummy-fn` (via `#'times`)")
      (is (= expected
             (xref/fn-transitive-deps dummy-fn))))))
```

test/orchard/xref_test.clj
vemv: > Not sure if this test has any value, given that times now appears in all the other tests.

Right, please craft a separate example such that it only a lambda dependency, so that it's super clear what we're testing.

i.e. it's not ideal to test many things at the same time.

As in a prior note, please avoid "lambda" terminology, in a deftest I'd suggesting a super clear wording so that we can maintain this effectively.

Finally, you can also rename `times` to something like `fn-dep` so that we emphasize what is being tested.

vemv (commented):
One final note from my side

vemv:
Cheers 🍻

Issue #65, opened 2019-07-17
Move ClojureScript dep to profiles


bbatsov:
Run now we've hardcoded the dep, which is bad for end users. It should go to all profiles the way we've done for other projects (e.g. https://github.com/clojure-emacs/cider-nrepl/blob/master/project.clj). 

@arichiardi The current code will work with older ClojureScript releases, right?

SevereOverfl0w:
I don't know the state of the code, but can we ensure that all the code works without cljs loaded, lest we end up with the same problem we have at piggieback.

bbatsov:
Well, that would be ideal, but it might be a bit harder here. When I agreed to baking the ClojureScript support in Orchard I didn't now many deps ClojureScript was pulling in (I assumed it didn't have any runtime deps). While that's not an issue for `cider-nrepl`, which obfuscates the deps, it will be an issue for everyone using `orchard` directly.

SevereOverfl0w:
The clojurescript dependency cannot be obfuscated. Otherwise you won't be accessing the right clojurescript state (via the compiler env)

bbatsov:
Ah, yeah. Silly me. Well, I guess we'll figure something out.

SevereOverfl0w:
Piggieback solved this with a "shim" to only load cljs code when cljs was available.

PR #143, opened 2022-01-08
Make ClojureScript dependency :provided

vemv:
Fixes https://github.com/clojure-emacs/orchard/issues/65


Makefile
```diff
@@ -9,13 +9,13 @@ curl -o $@ https://github.com/clojure-emacs/clojuredocs-export-edn/raw/master/ex
 
 # .EXPORT_ALL_VARIABLES passes TEST_PROFILES to Lein so that it can inspect the active profiles, which is needed for a complete Eastwood setup:
 test: clean .EXPORT_ALL_VARIABLES
-	lein with-profile -user,-dev,+$(VERSION),$(TEST_PROFILES) test
+	lein with-profile -user,-dev,+$(VERSION),$(TEST_PROFILES),-provided test
```
vemv, line 12: My thinking is that we should exercise the test suite with both `-provided` and no `-provided`, ensuring that the functionality works under any scenario.

Are you on board with the idea?

bbatsov: Yeah, that makes sense.

vemv:
Took me a few attempts, but it's ready+green now!

bbatsov (approved):

Issue #142, opened 2022-01-04
orchard.namespace/read-namespace fails if the ns form includes reader conditionals


alasdairm:
## Expected behavior
orchard.namespace/read-namespace should pull out the name from the first  (ns ..,) form in any cljc file.

## Actual behavior
orchard.namespace/read-namespace returns nil if the ns form includes a reader conditional.
(the EDN read throws an exception that is caught by the catch block).

## Steps to reproduce the problem
Can be reproduced in two ways - from CIDER or by adding a test to Orchard that (IMO) should pass (but fails).

### From CIDER
create a project with a mix of files with and without reader conditionals in their ns forms
E.g.
```
(ns my-namespace
  (:require #?(:clj [clojure.test :refer :all]
                     :cljs [cljs.test :as t :include-macros true])))
```
Select load all project files
Browse all namespaces
observe that the files with reader conditionals have not been loaded.

### By adding a test to Orchard
Add a test that passes a file with reader conditionals to read-namespace

```
--- a/test/orchard/namespace_test.clj
+++ b/test/orchard/namespace_test.clj
@@ -65,8 +65,13 @@
       (testing "of top-level forms only"
         (spit url "(comment (ns ns1)) (ns ns2) (ns ns3)")
         (is (= (n/read-namespace uri) 'ns2)))
+      (testing "of ns forms with reader conditionals"
+        (spit url "(ns ns1 (:require #?(:clj [clojure.test] :cljs [cljs.test])))")
+        (is (= 'ns1
+               (n/read-namespace uri))))
       (io/delete-file url))))

+
 (deftest namespace-resolution
   (testing "Resolving"
     (let [nses '[clojure.java.io
```
Run the test 
Error reported - read-namespace returns nil rather than ns1

## Environment & Version information
CIDER 1.2.0 (Nice), nREPL 0.9.0
refactor-nrepl/refactor-nrepl "3.1.0"
cider/cider-nrepl "0.27.4"
cider/orchard "0.8.0"

### Clojure version
Clojure 1.10.1

### Java version
Java 14.0.2

### Operating system
Windows 10

### Suggested fix [edited]
The problem is that clojure.edn/read throws an exception if it encounters a reader conditional. The exception is caught and nil returned. 

Instead of clojure.edn/read use clojure.core/read with *read-eval* bound to false and passing in an option of :read-cond :allow.

[...]

PR #145, opened 2022-01-08
Make read-namespace handle read conditionals gracefully

vemv:
Fixes https://github.com/clojure-emacs/orchard/issues/142

### DIFF
--- a/src/orchard/namespace.clj
+++ b/src/orchard/namespace.clj
@@ -6,7 +6,6 @@
   {:author "Jeff Valk"
    :added "0.5"}
   (:require
-   [clojure.edn :as edn]
    [clojure.java.io :as io]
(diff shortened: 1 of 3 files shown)


src/orchard/namespace.clj
```diff
@@ -20,13 +19,21 @@
 (defn read-namespace
   "Returns the namespace name from the first top-level `ns` form in the file."
   [url]
-  (try
-    (with-open [r (PushbackReader. (io/reader url))]
-      (->> (repeatedly #(edn/read r))
```
vemv, line 25: `repeatedly` + `filter` + clojure's chunked sequences would mean that we would perform more reads than necessary, which is detrimental to perf.

bbatsov:
Looks good to me.

PR #147, opened 2022-01-09
fixes AoT function lookup

### LINKED ISSUE #146: AoT compiled function dependencies don't show up
```clojure
(fn-deps reverse)
;; => #{}
;; expected:
;; => #{#clojure.core/reduce1 #clojure.core/conj}
```

This is because the AoT code is not loaded by Clojures `DynamicClassLoader`. 
I've got a fix for some of this in a bit.
The harder problem is that some functions, like `map` inherit from `clojure.lang.RestFn` and do not have references to their dependencies. I'm still investigating if we can somehow cover those.

Cyrik:
Fixes #146

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/orchard/xref.clj
+++ b/src/orchard/xref.clj
@@ -57,13 +57,18 @@
   {:added "0.5"}
   [v]
   (when-let [^clojure.lang.AFn v (to-fn v)]
-    (let [f-class-name (-> v .getClass .getName)]
-      ;; this uses the implementation detail that the clojure compiler always
-      ;; prefixes names of lambdas with the name of its surrounding function class
-      (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
-                      (map (fn [[_k value]] (.get ^java.lang.ref.Reference value)))
-                      (mapcat fn-deps-class))
-            class-cache))))
+    (let [f-class-name (-> v .getClass .getName)
+          ;; this uses the implementation detail that the clojure compiler always
+          ;; prefixes names of lambdas with the name of its surrounding function class
+          deps (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
+                               (map (fn [[_k value]] (.get ^java.lang.ref.Reference value)))
+                               (mapcat fn-deps-class))
+                     class-cache)]
+      ;; if there's no deps the class is most likely AoT compiled,
+      ;; try to access it directly
+      (if (empty? deps)
+        (-> v .getClass fn-deps-class)
+        deps))))
 
 (defn fn-transitive-deps
   "Returns a set with all the functions invoked inside `v` or inside those functions.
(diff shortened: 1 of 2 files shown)


src/orchard/xref.clj
```diff
@@ -57,13 +57,18 @@
...
+    (let [f-class-name (-> v .getClass .getName)
+          ;; this uses the implementation detail that the clojure compiler always
+          ;; prefixes names of lambdas with the name of its surrounding function class
+          deps (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
+                               (map (fn [[_k value]] (.get ^java.lang.ref.Reference value)))
+                               (mapcat fn-deps-class))
+                     class-cache)]
+      ;; if there's no deps the class is most likely AoT compiled,
+      ;; try to access it directly
+      (if (empty? deps)
```
vemv, line 69: What if it's just a defn with no deps?

e.g. some new code I'm just writing now.

Cyrik: then fn-deps-class will still return an empty set, we just waste some time by calling it twice.
the other option would be to run the filter outside the transducer to check if it returns empty. that would mean we don't have to run fn-deps-class twice for empty functions.
In the end, running fn-deps-class twice on empty functions doesn't really make much of a difference since it's very fast, especially when it doesn't have anything to return.

test/orchard/xref_test.clj
```diff
@@ -24,7 +24,10 @@
   (testing "with a symbol"
     (is (= (xref/fn-deps 'orchard.xref-test/dummy-fn)
```
vemv, line 25: Swap order of = hands please, normally it's `= <expected> <actual>`

src/orchard/xref.clj
vemv: Alright then!

test/orchard/xref_test.clj
```diff
@@ -52,7 +55,7 @@
   (testing "basics"
     (let [expected #{#'orchard.xref-test/fn-deps-test #'orchard.xref-test/fn-dep #'clojure.core/even?
                      #'clojure.core/filter #'orchard.xref-test/fn-transitive-dep #'clojure.core/map
-                     #'clojure.test/test-var #'clojure.core/range}]
+                     #'clojure.test/test-var #'clojure.core/range #'clojure.core/inc'}]
```
vemv, line 58: It would help me to understand, why is `inc'` returned now and not before?

Aren't both `#'clojure.core/range` and `#'clojure.core/inc'` aot-compiled?

Cyrik: That's one of the non-static functions in core, like `conj`. It gets called by `range` as you said, but only the 0 arity version. 
The interesting thing is that inc' has an `:inline` meta but because it doesn't call the function, just passes it to `iterate` it actually does it with the var.

There are two new learnings here: inline dependencies get found if they are just passed along, which does make sense and we always return all arity dependencies. I don't think we can fix the arity thing.

It might make sense to check in with Alex Miller at some point to see if he would be willing to make some modifications to the clojure compiler to make these reflection type things easier and more accurate.

vemv: Thanks!

Please simply add a more explicit test just like a past one:

image

so that we can remember the special meaning inc' has here.

vemv:
Thanks!

PR #149, opened 2022-01-17
Update namespace.clj

niquola:
io/resource may return nil in some situations (in my case it was emacs recovery file)

### DIFF
--- a/src/orchard/namespace.clj
+++ b/src/orchard/namespace.clj
@@ -112,9 +112,9 @@
         (apply concat)
         (pmap (fn [x]
                 (when (misc/clj-file? x)
-                  x)))
+                  (io/resource x))))
         (filter identity)
-        (pmap (comp read-namespace io/resource))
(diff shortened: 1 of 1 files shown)


vemv:
Hi! Thanks fofr the contribution. Could you describe the problem being solved?

Thanks - V

bbatsov:
Some [background](https://twitter.com/niquola/status/1483011905681825795?s=20). I'm also not sure when exactly this problem manifests. We'll need some unit tests covering this scenario as well.

vemv:
In any case, maybe the fix could be moved to `misc/clj-file?`

I'll have a calmer look at the screenshots

vemv:
I pushed the fix directly to master out of muscle memory, sorry.

It's essentially the same, with some bonus performance considerations.

If you can pass us an exact `emacs recovery file`, we'd be happy to add it to the test suite.

PR #154, opened 2022-02-22
Optimize `member-info`

vemv:
* Optimize `member-info`
  * Its usage of laziness could likely mean that more objects would be retained than necessary, causing OOMs in constrainted environments.
  * Reported in https://clojurians.slack.com/archives/C0617A8PQ/p1645530472596039
* Bump enrich-classpath
* Bump Eastwood


src/orchard/java.clj
```diff
@@ -268,11 +268,13 @@
   implemention. If the member is an instance member, `this` is prepended to its
   arglists."
   [class member]
-  (let [c (->> (class-info class)
-               (iterate (comp class-info :super))
-               (take-while identity)
-               (filter #(get-in % [:members member]))
-               (first))]
+  (let [c (loop [next-class class]
```
bbatsov, line 271: As this code is not idiomatic it should definitely include some comments to discourage people from optimizing it in the other direction. :-)

vemv:
Thanks for the review!

PR #159, opened 2022-05-09
Make orchard.namespace/classpath-namespaces resilient to faulty ns declarations

### LINKED ISSUE #158: error reporting fails when project contains invalid namespace declaration
## Expected behavior
I expect orchard `orchard.namespace/classpath-namespaces` to be resilient to faulty ns declarations.

## Actual behavior
The precence of a faulty ns declaration on the class path like `(ns (:requires [clojure.string]))` causes nrepl to swallow the error and just print an exception to the REPL output.

``` 
git clone https://gist.github.com/mk/bd82092048d733ef3de0e312898f47b4 nrepl-issue
```

Open the project and run `cider-jack-in`

Try evaluating any expression that throws an expression (there's one in foo.clj).

### Clojure version
1.11.1

### Java version
openjdk version "17.0.3" 2022-04-19
OpenJDK Runtime Environment GraalVM CE 22.2.0-dev (build 17.0.3+4-jvmci-22.1-b03)
OpenJDK 64-Bit Server VM GraalVM CE 22.2.0-dev (build 17.0.3+4-jvmci-22.1-b03, mixed mode, sharing)

### Operating system
macOS Monterey

I'll follow up with a PR shortly.

mk:
Closes #158

- [x] All tests are passing

### DIFF
--- a/test/orchard/namespace_test.clj
+++ b/test/orchard/namespace_test.clj
@@ -62,6 +62,9 @@
       (testing "when multiple ns forms are present"
         (spit url "(ns ns1) (ns ns2) (ns ns3)")
         (is (= 'ns1 (sut/read-namespace uri))))
+      (testing "when ns form is invalid"
+        (spit url "(ns (:require [clojure.string]))")
+        (is (nil? (sut/read-namespace uri))))
       (testing "of top-level forms only"
         (spit url "(comment (ns ns1)) (ns ns2) (ns ns3)")
         (is (= 'ns2 (sut/read-namespace uri))))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## master (unreleased)
+* [#158](https://github.com/clojure-emacs/orchard/issues/158): Make classpath-namespaces resilient to faulty ns declarations.
 
 ## 0.9.2 (2022-02-22)
 
--- a/src/orchard/namespace.clj
+++ b/src/orchard/namespace.clj
@@ -31,7 +31,8 @@
         (cond
           (#{::eof ::fail} found) nil
           (and (list? found)
-               (-> found first #{`ns 'ns}))
+               (-> found first #{`ns 'ns})
+               (symbol? (second found)))
           (second found)
           :else (recur))))))
 


test/orchard/namespace_test.clj
```diff
@@ -62,6 +62,9 @@
       (testing "when multiple ns forms are present"
         (spit url "(ns ns1) (ns ns2) (ns ns3)")
         (is (= 'ns1 (sut/read-namespace uri))))
+      (testing "when ns form is invalid"
+        (spit url "(ns (:require [clojure.string]))")
```
vemv, line 66: `(ns foo (:requires [clojure.string]))` is a also a good example to be covered.

Accordingly, probably a try/catch would be better than the proposed approach - there can be few reasons why a ns parsing could fail.

vemv (commented):
I think the proposal SGTM, thanks!

Perhaps we were intentionally encouraging that namespaces would be correct at every moment - it's part of CIDER's approach to have loadable code that can be _run_ (and not only parsed).

But it's always possible that some 3rd party library could accidentally include some faulty unused namespace, which isn't something users can easily fix. So it's more friendly to give users something usable, than not.

test/orchard/namespace_test.clj
mk: The error I've reported in #158 does not occur here, but it originates here, see:

https://github.com/clojure-emacs/orchard/blob/c0e24f71292f4fed8d4d8cf854a99e4a80cf6ee6/src/orchard/namespace.clj#L121-L123

What happens without my change is that we try to sort and hence compare a seq like `(:requires [clojure.string])` with symbols and that fails.

I do think a try/catch could be good & helpful in addition to that, but then I'd place it way further up the stack, possibly in cider-nrepl in `analyze-stacktrace` https://github.com/clojure-emacs/cider-nrepl/blob/5c0f21197fcccb1b2ca67054cab1dcc8a6af2c7f/src/cider/nrepl/middleware/stacktrace.clj#L216

Btw, re-reading the issue I'm noticing now I should have pointed out the stacktrace in https://gist.github.com/mk/bd82092048d733ef3de0e312898f47b4 more promintely.

mk:
> Perhaps we were intentionally encouraging that namespaces would be correct at every moment - it's part of CIDER's approach to have loadable code that can be run (and not only parsed).

Looking at the tests for `read-namespace` makes me think it's built to namespaces that cannot be read:
https://github.com/clojure-emacs/orchard/blob/c0e24f71292f4fed8d4d8cf854a99e4a80cf6ee6/test/orchard/namespace_test.clj#L53-L61

> But it's always possible that some 3rd party library could accidentally include some faulty unused namespace, which isn't something users can easily fix. So it's more friendly to give users something usable, than not.

I ran into two cases on two different project in the past week were this assumption broke. It was faulty code under our control but we didn't notice what was causing the issue or where to look for an error.

In addition to this chance I think it makes sense to be more defensive in `cider-nrepl` and catch exceptions thrown in the stacktrace handling and warn the user about what's wrong. Let me know if you think this is a good idea and I could take a look at that as a follow-up.

vemv:
Thanks for the insights!

> In addition to this chance I think it makes sense to be more defensive in cider-nrepl

I also think so, and would prefer that we think that we handle all missing edge cases in a single PR - else it's just "git churn" that we're creating (i.e. make a change, then undo it to do something more comprehensive).

vemv:
Alright, respected, I'll give a look at the problem as I find a chance asap.

jackrusher:
Delaying this PR to wait for the final comprehensive fix reminds me of:
https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good

vemv:
It's pretty safe to assume that most OSS maintainers are already familiar with various truisms (and their applicability to specific scenarios).

bbatsov:
So, are we going to do something with this PR or not?

vemv (approved):
> So, are we going to do something with this PR or not?

I checked out this PR again the other day and it seemed fine to me with no future tweaks needed, since we can rely on the compiler for checking unloadable code.

bbatsov:
Good. I'm planning a new CIDER release (perhaps as soon as this week) and it'd be nice if we manage to squeeze in recent Orchard updates there.

PR #168, opened 2023-07-19
Upgrade misc dependencies

vemv:
* Upgrade misc dependencies
* Expand CI matrix

Cheers - V


bbatsov:
@vemv From the CI only on drop the support for it? Generally we can't drop the support for 1.8 before cutting the last CIDER 1.x release.

bbatsov (approved):

PR #169, opened 2023-07-19
Introduce `orchard.indent`

vemv:
Part of https://github.com/clojure-emacs/cider-nrepl/issues/777

Cheers - V

### DIFF
--- a/src/orchard/indent.clj
+++ b/src/orchard/indent.clj
@@ -2,8 +2,7 @@
   "`:style/indent` inference."
   (:require
    [clojure.set :as set]
-   [clojure.string :as string]
-   [clojure.walk :as walk])
+   [clojure.string :as string])
   (:import
    (java.util List)))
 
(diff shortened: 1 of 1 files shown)


src/orchard/indent.clj
```diff
@@ -0,0 +1,173 @@
...
+                   clojure.core/proxy        [2 nil nil [:defn]]
+                   clojure.core/reify        [:defn [1]]
+                   clojure.test/testing      1
+                   this-as                   1
+                   clojure.core.async/thread 0
+                   try                       0
+                   clojure.core/when         1
+                   clojure.core/while        1}))
+
+(def clojure-mode-indents-fuzzy
```
bbatsov, line 50: I don't get the need for fuzzy matches, as I think `clojure-mode` only does exact matches. Can you elaborate on this?

bbatsov: That's fair, but let's just document this better.

bbatsov:
Fine by me. For the record - I was thinking more about the user docs (CIDER's docs) on indentation.

PR #170, opened 2023-08-01
[inspector] Fix exception thrown when inspecting short Eduction

alexander-yakushev:
Because `Eduction` type implements `clojure.lang.Sequential` but not `clojure.lang.Counted`, the inspector threw an exception when trying to inspect it. A minor but annoying bug. Besides, there could be other similar classes out there, so it's worth fixing.

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -372,14 +372,12 @@
        ins))))
 
 (defn- last-page [{:keys [current-page page-size]} obj]
-  (if (or (instance? clojure.lang.Counted obj)
-          ;; if there are no more items after the current page,
-          ;; we must have reached the end of the collection, so
-          ;; it's not infinite.
-          (empty? (drop (* (inc current-page) page-size) obj)))
-    (quot (dec (count obj)) page-size)
-    ;; possibly infinite
-    Integer/MAX_VALUE))
+  (cond (instance? clojure.lang.Counted obj) (quot (dec (count obj)) page-size)
+        ;; if there are no more items after the current page, we must have
+        ;; reached the end of the collection, so it's not infinite.
+        (empty? (drop (* (inc current-page) page-size) obj)) current-page
+        ;; possibly infinite
+        :else Integer/MAX_VALUE))
 
 (defn- render-page-info [{:keys [current-page page-size] :as inspector} obj]
   (if-not (sequential? obj)
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -379,8 +379,10 @@
         ;; possibly infinite
         :else Integer/MAX_VALUE))
 
+(declare known-types)
+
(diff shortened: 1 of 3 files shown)


src/orchard/inspect.clj
```diff
@@ -372,14 +372,12 @@
...
-          (empty? (drop (* (inc current-page) page-size) obj)))
-    (quot (dec (count obj)) page-size)
-    ;; possibly infinite
-    Integer/MAX_VALUE))
+  (cond (instance? clojure.lang.Counted obj) (quot (dec (count obj)) page-size)
+        ;; if there are no more items after the current page, we must have
+        ;; reached the end of the collection, so it's not infinite.
+        (empty? (drop (* (inc current-page) page-size) obj)) current-page
+        ;; possibly infinite
+        :else Integer/MAX_VALUE)) cat
```
vemv, line 380: `cat`? 🐱

test/orchard/inspect_test.clj
```diff
@@ -897,6 +897,21 @@
                           (:newline)))
                       (datafy-section rendered))))))))
 
+(deftest inspect-eduction-test
+  (testing "inspecting eduction shows its obejct fields"
+    (let [rendered (-> (eduction (range 10)) inspect render)]
+      (testing "renders the header section"
+        (is (match? '("Class"
+                      ": "
+                      (:value "clojure.core.Eduction" 0)
```
vemv, line 906: Probably worth playing with `inspect/next-page` as other tests do

vemv (approved):
lgtm!

src/orchard/inspect.clj
alexander-yakushev: Useless use of cat! https://porkmail.org/era/unix/award

test/orchard/inspect_test.clj
alexander-yakushev: It's interaction with paging is weird right now (pager shows like it's paged but it's not, it just shows Java fields). See my comment below.

alexander-yakushev: Actually, fixed that weird interaction too since it was ugly.

vemv:
Tests are failing on Clojure master

Besides from that, all lgtm, if you want to expand the tests you are on time since I'll create an unrelated PR tomorrow, so it will get all in the same release train.

alexander-yakushev:
Hmm, looks like a flaky test to me, I don't think I changed anything that could trigger that case.

alexander-yakushev:
See, passed without changes:)

vemv:
🍻!

With a bit of luck we'll have new orchard/cider-nrepl today/tomorrow

alexander-yakushev:
🍎 🍎 🍎 
🍺 🍺 🍺 
🥴 🥴 🥴

vemv:
https://clojars.org/cider/cider-nrepl/versions/0.34.0

PR #171, opened 2023-08-02
Infer var metadata for 'proxy' vars

vemv:
Part of https://github.com/clojure-emacs/cider-nrepl/issues/788

Use case explained over there.

...I had attempted this before: https://github.com/clojure-emacs/cider-nrepl/pull/790 but it was targeting the wrong middleware. This impl is simpler / more predictable, needing no `pmap`-like parallelism.

I've tried this one locally, and it works as intended.

Works on CLJ and CLJS, tests added for both.

Cheers - V

### DIFF
--- a/src/orchard/cljs/analysis.cljc
+++ b/src/orchard/cljs/analysis.cljc
@@ -2,9 +2,10 @@
   "ClojureScript analysis functions."
   {:author "Gary Trakhman"
    :added "0.5"}
+  (:refer-clojure :exclude [all-ns find-ns find-var ns-aliases ns-resolve])
   (:require
-   [orchard.misc :as misc])
-  (:refer-clojure :exclude [find-ns find-var all-ns ns-aliases]))
+   [clojure.string :as string]
+   [orchard.misc :as misc]))
 
 (defn all-ns [{namespaces :cljs.analyzer/namespaces}]
   (into {}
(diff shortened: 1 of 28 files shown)


bbatsov:
Is `proxy/proxied` the right terminology here? I think `alias/aliased` is much more accurate way to describe this.

vemv:
I can see how 'proxy' is an overloaded term, but 'aliased' immediately brings to mind namespace aliases, which would seem greatly confusing to me (particularly in a project that already deals with aliases).

`delegate` might work, but I'd fail to see a difference with `proxy`.

bbatsov:
When I hear of Clojure and proxy I immediately think of https://clojuredocs.org/clojure.core/proxy That's why I don't like this. And your examples in the tests are really aliases, after all this terminology extends to more than namespaces. 

I'm open to other suggestions, even if I can't offer anything better than `alias` myself. @alexander-yakushev Any suggestions from you are welcome as well!

src/orchard/cljs/analysis.cljc
```diff
@@ -2,9 +2,10 @@
   "ClojureScript analysis functions."
   {:author "Gary Trakhman"
    :added "0.5"}
+  (:refer-clojure :exclude [all-ns find-ns find-var ns-aliases ns-resolve])
   (:require
-   [orchard.misc :as misc])
-  (:refer-clojure :exclude [find-ns find-var all-ns ns-aliases]))
+   [clojure.string :as string]
```
alexander-yakushev, line 7: It seems that `str` is the preferred alias across the project, any reason to have it different here?

bbatsov:
Btw, I don't want us to spend too much time on this - the implementation seems good and solves a real problem. If we don't come up with a better name relatively quickly (a day or so), I'm fine to stick with the current terminology.

src/orchard/info.clj
```diff
@@ -98,37 +98,47 @@
      ;; a special symbol - always use :unqualified-sym
      (some-> (cljs-ana/special-meta env unqualified-sym)
              (cljs-meta/normalize-var-meta))
-     ;; an NS
+
+     ;; a NS
```
alexander-yakushev, line 102: Well ackhually, `an NS` is correct.

alexander-yakushev:
> When I hear of Clojure and proxy I immediately think of https://clojuredocs.org/clojure.core/proxy That's why I don't like this. And your examples in the tests are really aliases, after all this terminology extends to more than namespaces.

I absolutely agree; I was reading through the code waiting and waiting where a `proxy` finally comes up, and only understood what it is about when looked at the tests.

I think it's worth picking up a better term. `alias` has its confusion too but still better than `proxy`. Maybe something like "indirect"? "Pass-through"?

src/orchard/cljs/analysis.cljc
bbatsov: Agreed - it's best to be consistent about this.

src/orchard/info.clj
bbatsov: Indeed, it's pronounced as `en es`, so the normal rules about consonants don't apply here.

src/orchard/cljs/analysis.cljc
vemv: If you don't mind I'd like to settle on`string` then, project-wide. We can add a kondo rule for avoiding further churn.

My point of view as someone who tends to maintain other people's code a lot, is that Sierra-style aliases only make our lives easier.

bbatsov: It's not a big deal for me, but such changes should be discussed outside the scope of this PR.

vemv: For me, most of all it's about not memorizing arbitrary abbreviations someone wrote years ago e.g. `cljs.analyzer.api :as ana` - those represent unnecessary cognitive load to me.

I agree that `str` is very well-known, but when one uses a convention 99.9% of the time, having an exception to break the rule isn't exactly convenient.

alexander-yakushev: So then everyone will have to memorize whether this is a project that chose `string` because of a convention, or you know the rest 99.9% of the projects 😄 .

vemv:
Merging. btw, I tested out this with Compliment and it works there too 🍻

Issue #173, opened 2023-08-04
"fn-refs" op is slow


vemv:
I found that fn-refs takes a few seconds to complete. cider.el calls this synchronously:

```clj
cider.nrepl.middleware.xref> (time
                               (cider.nrepl.middleware.xref/fn-refs-reply {:ns "orchard.misc"
                                                                           :sym "as-sym"}))
"Elapsed time: 3758.845921 msecs"
{:fn-refs
 ({:name "cider.nrepl.middleware.xref/fn-refs-reply",
   :doc "(not documented)",
   :file "cider/nrepl/middleware/xref.clj",
   :line 19,
   :column 1}
,,,
```

(fn-deps does not have this slowness)

Maybe there a few aspects to optimize?

PR #174, opened 2023-08-04
Parallelize `orchard.xref/fn-refs`

vemv:
Fixes https://github.com/clojure-emacs/orchard/issues/173

#### Before

```clj
(time
  (do
    (doall (xref/fn-refs #'orchard.misc/as-sym))
    1))
"Elapsed time: 3570.653465 msecs"
```

#### After

```clj
(time
  (do
    (doall (xref/fn-refs #'orchard.misc/as-sym))
    1))
"Elapsed time: 528.020007 msecs"
```

`pmap` isn't exactly the best. I had a nice Claypoole integration in a discarded PR recently. Can be a global improvement after discussion later.

Cheers - V

### DIFF
--- a/src/orchard/xref.clj
+++ b/src/orchard/xref.clj
@@ -19,6 +19,10 @@
 (defn- fn-name [^java.lang.Class f]
   (-> f .getName repl/demunge symbol))
 
+(def eval-lock
+  "We don't want parallel evaluation - easily dangerous."
+  (Object.))
+
 (defn fn-deps-class
   "Returns a set with all the functions invoked by `v`.
   `v` can be a function class or a symbol."
@@ -26,7 +30,8 @@
   [v]
   (let [^java.lang.Class v (if (class? v)
                              v
-                             (eval v))]
+                             (locking eval-lock
+                               (eval v)))]
     (into #{} (keep (fn [^java.lang.reflect.Field f]
                       (or (and (identical? clojure.lang.Var (.getType f))
                                (java.lang.reflect.Modifier/isPublic (.getModifiers f))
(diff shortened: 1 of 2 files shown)


bbatsov:
> pmap isn't exactly the best. 

I know this, but I strongly favor avoid dependencies in Orchard. `pmap` might not be perfect, but it gets the job done.

src/orchard/xref.clj
```diff
@@ -60,9 +65,12 @@
     (let [f-class-name (-> v .getClass .getName)
           ;; this uses the implementation detail that the clojure compiler always
           ;; prefixes names of lambdas with the name of its surrounding function class
-          deps (into #{} (comp (filter (fn [[k _v]] (clojure.string/includes? k f-class-name)))
```
bbatsov, line 63: The whitespace changes add some noise. I prefer to those in separate commits, so it's clear what's the meaningful part of the change and what is simply stylistic.

vemv: Done. Will keep in mind for next times.

vemv (commented):
> I know this, but I strongly favor avoid dependencies in Orchard. pmap might not be perfect, but it gets the job done.

Maybe at some point we can offer our own `orchard.parallel` ns with a handy pmap.

After all, Claypool does nothing that cannot be done with a basic understanding of executors.

bbatsov:
> > I know this, but I strongly favor avoid dependencies in Orchard. pmap might not be perfect, but it gets the job done.
> 
> Maybe at some point we can offer our own `orchard.parallel` ns with a handy pmap.
> 
> After all, Claypool does nothing that cannot be done with a basic understanding of executors.

I like this idea!

PR #184, opened 2023-09-02
`info` cljs: don't mistakenly prioritize special form names over var names

### LINKED ISSUE #182: `info` returns the wrong docstring for cljs `spec/def`
`spec/def` is a macro.

For this cljs request, the docstring for the special form `def` is returned instead of `spec/def`'s.

```clj
(-->
  id         "51"
  op         "eldoc"
  session    "acfd603d-57fc-41a6-bb79-38c192d18fa2"
  time-stamp "2023-08-29 13:41:49.282307000"
  ns         #("icd.scroll.letters" 0 18 (fontified t cider-locals nil cider-block-dynamic-font-lock t face font-lock-type-face))
  sym        "spec/def"
)
(<--
  id         "51"
  session    "acfd603d-57fc-41a6-bb79-38c192d18fa2"
  time-stamp "2023-08-29 13:41:49.284298000"
  docstring  "Creates and interns a global var with the name
  of symbol in the current namespace (*ns*) or locates such a var if
  it already exists.  If init is supplied, it is evaluated, and the
  root binding of the var is set to the resulting value.  If init is
  not supplied, the root binding of the var is unaffected."
  eldoc      (("def" "symbol" "doc-string?" "init?"))
  name       "def"
  ns         "cljs.core"
  status     ("done")
  type       "special-form"
)
```

vemv:
Fixes #182

Cheers - V

### DIFF
--- a/test/orchard/info_test.clj
+++ b/test/orchard/info_test.clj
@@ -1,6 +1,7 @@
 (ns orchard.info-test
   (:require
    [clojure.java.io :as io]
+   [clojure.set :as set]
    [clojure.string :as string :refer [replace-first]]
    [clojure.test :refer [are deftest is testing use-fixtures]]
    [orchard.info :as info]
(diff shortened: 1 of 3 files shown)


test/orchard/info_test.clj
```diff
@@ -99,24 +100,30 @@
...
+                          (let [{specials true} (->> info-params
+                                                     (map info/info*)
+                                                     (group-by :special-form))]
+                            (into #{}
+                                  (map :name)
+                                  specials)))]
       (when cljs-available?
         (testing "- :cljs"
           (let [special-doc-map (misc/require-and-resolve 'cljs.repl/special-doc-map)
-                special-syms (into '#{in-ns load load-file} (keys special-doc-map))]
```
vemv, line 110: I removed `load-file` because it's a public function (https://github.com/clojure/clojurescript/blob/17068fa91cae169ed7c881adb593d85138efe6d0/src/main/cljs/cljs/core.cljs#L338) _and_ a special form (https://cljs.github.io/api/cljs.core/load-file)

Because that's such an exceptional case, for a rarely-used piece of functionality, I didn't spend much time deliberating whether to favor one or the other.

Issue #179, opened 2023-08-20
Improve handling of html within Java docstrings


vemv:
Given this Java docstring:

<img width="789" alt="image" src="https://github.com/clojure-emacs/orchard/assets/1162994/acab048d-ff00-48c2-be51-fdae75535cae">

I see it offered like this in Compliment (which Orchard's `info` is responsible for):

<img width="790" alt="image" src="https://github.com/clojure-emacs/orchard/assets/1162994/6fc54021-8cfc-4ffc-a2f9-12fef4e63ecb">

...I'm guessing that we try to translate the html table to an 'ascii' table. But without alignment it looks rather confusing.

CIDER side, we also could try to colorize the table delimiters somehow.

n.b. there might be a touch of 'personal setup' factor here, although I believe I'm seeing stock behavior.

vemv:
As seen in the screenshot, we also can replace `{@code ...}` with simple backticks.

PR #185, opened 2023-09-09
Rework java `:doc`

vemv:
> Fixes https://github.com/clojure-emacs/orchard/issues/179

## Context

For Java stuff, we return `:doc` which is the parsed doc comments form the Java sources.

They're originally in a html-like markup language.

The existing approach parsed that html-like text with a HTML parser. Then, the HTML was converted to Markdown, which certainly does read more nicely.

However, that approach had a number of flaws:

* Newlines (not expressed as `<br>`, e.g. `\n`) were lost
  * I find this flaw critical - without newlines, docstrings won't look nice in a UI
    * e.g. eldoc trimming won't work properly.
  * There's no workaround for this, since the HTML parser we were using eats newlines before we can do anything with them.
* Tables were not aligned, which hindered their readability
* non-html bits like `@code`, `@since`, etc were not parsed at all
  * since that isn't html, the html parser could not help us at all, and would demand further bespoke solutions.

This is the state of things:

image

## Proposed changes

* Stop parsing converting html to markdown
* Return the original doc comments under `:doc` - they're of interest (many IDEs _can_ render html) and a fallback if our alternatives fail.
* Parse the doc comments using the official 'doc comment' parser that is bundled with newer JDKs.
* Include two new keys:  `:doc-fragments`, `:doc-first-sentence-fragments`
  * Both represent the same thing - the latter is limited to the first sentence
  * Example value: `[{:content "Class level docstring.\n\n " :type "text"} {:content "<pre> foo </pre>", :type "html"}]`
  * These represent a succession (vector) of html and non-html content
  * Orchard clients can take this succession, and render the html parts with a renderer of choice, and then join all the fragments separated by a newline.

This is how things are looking after integrating my solution into cider-nrepl and cider.el:

image (6)

...I render the html fragments with [shr.el](https://github.com/emacs-mirror/emacs/blob/01ba4d6c5f9622ce861b53760ddd05b788e27c95/lisp/net/shr.el), which gives us table alignment, italics, bold, etc.

One also can custom-render things. For instance I can tell shr to render `<pre>` blocks with java-mode:

image (7)

Cheers - V

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -23,7 +23,9 @@
                                     :password :env/clojars_password
                                     :sign-releases false}]]
 
-  :jvm-opts ["-Dclojure.main.report=stderr"]
+  :jvm-opts ["-Dclojure.main.report=stderr"
+             "--add-opens"
+             "jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED"]
 
   :source-paths ["src" "src-jdk8" "src-newer-jdks"]
   :test-paths ~(cond-> ["test"]
--- a/project.clj
+++ b/project.clj
@@ -60,6 +60,8 @@
                                "-Dorchard.internal.has-enriched-classpath=false"]
                     :source-paths ["test" "src-spec-alpha-2/src/main/clojure"]}
 
+             :mark-enriched-classpath {:jvm-opts ["-Dorchard.internal.has-enriched-classpath=true"]}
+
              :enrich-classpath {:plugins [[mx.cider/enrich-classpath "1.17.0"]]
                                 :middleware [cider.enrich-classpath/middleware]
                                 :jvm-opts ["-Dorchard.internal.has-enriched-classpath=true"]
--- a/project.clj
+++ b/project.clj
@@ -55,9 +55,12 @@
                                      "does-not-exist.jar"]
                     :java-source-paths ["test-java"]
                     ;; Initialize the cache verbosely, as usual, so that possible issues can be more easily diagnosed:
-                    :jvm-opts ["-Dorchard.initialize-cache.silent=false"
-                               "-Dorchard.internal.test-suite-running=true"
-                               "-Dorchard.internal.has-enriched-classpath=false"]
+                    :jvm-opts
+                    ~(cond-> ["-Dorchard.initialize-cache.silent=false"
+                              "-Dorchard.internal.test-suite-running=true"
+                              "-Dorchard.internal.has-enriched-classpath=false"]
+                       (not jdk8?) (conj "--add-opens=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED"))
+
                     :source-paths ["test" "src-spec-alpha-2/src/main/clojure"]}
 
              :mark-enriched-classpath {:jvm-opts ["-Dorchard.internal.has-enriched-classpath=true"]}
--- a/project.clj
+++ b/project.clj
@@ -23,9 +23,7 @@
                                     :password :env/clojars_password
                                     :sign-releases false}]]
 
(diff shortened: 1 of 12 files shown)


project.clj
```diff
@@ -23,7 +23,9 @@
                                     :password :env/clojars_password
                                     :sign-releases false}]]
 
-  :jvm-opts ["-Dclojure.main.report=stderr"]
+  :jvm-opts ~(cond-> ["-Dclojure.main.report=stderr"]
+               (not jdk8?) (conj "--add-opens"
+                                 "jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED"))
```
vemv, line 28: We need this flag in order to the doc comment parser to be available.

enrich-classpath will add it, soon.

bbatsov:
I'll take a closer look tomorrow, but overall your idea seems sound.

src-newer-jdks/orchard/java/parser_next.clj
```diff
@@ -0,0 +1,259 @@
+(ns orchard.java.parser-next
+  "Source and docstring info for Java classes and members"
```
bbatsov, line 2: I'd explain here the differences with the other parser.

src-newer-jdks/orchard/java/parser_next.clj
```diff
@@ -0,0 +1,259 @@
...
+                            (reduce node-reducer node-reducer-init)
+                            :result)
+        first-sentence (some->> comment-tree
+                                .getFirstSentence
+                                (reduce node-reducer node-reducer-init)
+                                :result)]
+    {:doc (some-> env .getElementUtils (.getDocComment e))
+     :doc-first-sentence-fragments (coalesce first-sentence)
+     :doc-fragments (coalesce (into full-body block-tags))}))
+
```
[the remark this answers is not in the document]
bbatsov: `doc-first-sentence-fragments` sounds a bit weird, if this that's just the first sentence or a part of it. Might be good to document this structure in the code, so it's clear what is what exactly.

vemv:
@bbatsov : I've progressed another good chunk.

I'll add/fix tests tomorrow, but feel free to have one last look.

src-newer-jdks/orchard/java/parser_next.clj
```diff
@@ -0,0 +1,299 @@
+(ns orchard.java.parser-next
```
bbatsov, line 1: You should also mention in the metadata when was this ns added.

src-newer-jdks/orchard/java/parser_utils.clj
```diff
@@ -0,0 +1,113 @@
+(ns orchard.java.parser-utils
```
bbatsov, line 1: This needs a dostring and the added metadata.

src/orchard/java.clj
```diff
@@ -46,7 +46,7 @@
   "The `tools.jar` path, for JDK8 and earlier. If found on the existing
   classpath, this is the corresponding classpath entry. Otherwise, if available,
   this is added to the classpath."
-  (when (<= misc/java-api-version 8)
+  (when-not (>= misc/java-api-version 9)
```
bbatsov, line 49: I don't see the point in this change.

bbatsov:
Other than my small remarks it looks good to me.

src-newer-jdks/orchard/java/parser_next.clj
vemv: It wasn't intentional, before I changed the code to something else and then I 'reverted' it.

Anyway, the related conditions in this ns now look more homogeneous now.

Issue #186, opened 2023-09-13
Remove internal cache atoms


vemv:
We have some atoms, for caching.

It's likely better to leave things slower (for now) so that we don't favor OOMs.

See https://chat.openai.com/share/3069aebe-87ad-4354-a839-5c29f864f9e2

bbatsov:
Depends on how likely those OOMs are. I'm guessing the bulk of the users never experience them, given that I've rarely seen such reports.

vemv:
They support features that are only visible when using enrich-classpath, so most people haven't exercised those code paths a lot.

There's already some caution around memory, e.g. https://github.com/clojure-emacs/orchard/blob/c1ce92703f11c69229c90d70c95fee9e9ddabb61/src/orchard/java.clj#L262-L270

PR #187, opened 2023-09-19
Optimize internal cache

vemv:
> Fixes https://github.com/clojure-emacs/orchard/issues/186

Replaces the atom cache that would easily grow to huge values with a `LruCache` with capacity 250.

We also mostly stop populating the cache in advance. It's fairly pointless if we use a bounded capacity - no random classes are more important than others.

(We do populate the cache for a few key frequently-used classes, and whatever underlying classes they also need)

Cheers - V

### DIFF
--- a/java/mx/cider/orchard/LruMap.java
+++ b/java/mx/cider/orchard/LruMap.java
@@ -8,8 +8,9 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.concurrent.locks.ReadWriteLock;
 
 /**
- * A <code>java.util.Map</code> with a capacity expressed as <code>max_size</code>
+ * A <code>java.util.Map</code> with a capacity expressed as <code>max_size</code>.
  *
+ * It's meant to be used as a cache with Least Recently Used eviction policy:
  * When new <code>.put</code>s would surpass the capacity, older entries are discarded.
  *
  * This class is thread-safe.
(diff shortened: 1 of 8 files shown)


src/orchard/java.clj
```diff
@@ -223,14 +225,16 @@
 ;; and recompiled, we cache such classes with last-modified property, so that we
 ;; know when to purge those classes from cache.
 
-(def cache (atom {}))
+;; 250 is large enough to be useful (and hold a few key classes, like Object),
+;; and small enough to not incur into OOMs.
+(def ^Map cache (LruMap. 250))
```
bbatsov, line 230: I'm guessing that even double the size it wouldn't be an issue. You might want to expand the documentation a little bit to elaborate that this cache size is bounded and that's the reason why we went with a custom implementation.

bbatsov:
Overall it looks OK to me. Generally I avoid introducing Java code when possible, but in this case it's fairly simple so not a big deal. Alternatively you could have done something like https://github.com/clojure/core.cache/blob/master/src/main/clojure/clojure/core/cache.clj#L211

java/mx/cider/orchard/LruMap.java
```diff
@@ -0,0 +1,207 @@
...
+
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.locks.ReadWriteLock;
+
+/**
+ * A <code>java.util.Map</code> with a capacity expressed as <code>max_size</code>
```
bbatsov, line 11: You might mention that LRU stands for least recently used. :-) It took me a moment to remember this when I saw the class name.

vemv:
Thanks for the review!

Issue #190, opened 2023-09-22
Java arglists: include type metadata


vemv:
For a given Java arglist (as used in `*cider-doc*`, etc) like `[this key]`, we can also include:

* the return type
* the type of each param (except that of `this`)

So the example would become: `^Integer [this ^String key]`

classes other than `java.lang`'s should be fully-qualified.

Pay attention to primitives.

Disregard generics.

PR #192, opened 2023-09-26
Java source info: return `:formatted-arglists` when possible

vemv:
> Fixes https://github.com/clojure-emacs/orchard/issues/190

Performs some necessary preliminary fixes and then implements #190.

* Correctly consolidate arities info from two different sources (the Clojure Reflector, and .java source analysis) for a meaningful deep merge between then by computing `:non-generic-argtypes`
  * Without arities that match 1:1 between the two sources, we cannot deep merge correctly: there will be duplicate entries (given the slightly different keys), which results in extra (and incorrect) entries shown to end users.
  * This was a subtle bug in hidden in Orchard.
* Remove constructor info for now, given it's was incorrect and incomplete (and unused from cider.el) for both Reflector and Sources
  * e.g. for a given constructor with 4 different signatures, maybe just 1 was being shown, so it wasn't reliable enough.
  * See also https://github.com/clojure-emacs/orchard/pull/87 for evidence of constructors being a pending spot
  * Will be re-added soon enough
* Implement and return `:formatted-arglists` for CIDER usage

A few examples of `:formatted-arglists`:

```
^void [this, ^function.BiConsumer action]
^void [^int begin, ^int end, ^int length]
^Map$Entry [^Object k, ^Object v]
^StackTraceElement[][] [^Thread threads]
```

Showing `^StackTraceElement[][] [^Thread threads]` is far more useful to display to end users than merely `[threads]`.

A larger corpus of samples: https://gist.github.com/vemv/45c04bf8cd794c0902077f7db9cd5f3a

Cheers - V

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,7 +176,7 @@ jobs:
           steps:
             - run:
                 name: Running tests
-                command: ENRICH_CLASSPATH=<< parameters.enrich_classpath >> make --stop test
+                command: ENRICH_CLASSPATH=<< parameters.enrich_classpath >> make test
             # Eastwood is run for every item in the CI matrix, because its results are sensitive to the code in the runtime,
             # so we make the most out of this linter by exercising all profiles, JDK versions, etc.
             - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,7 @@ commands:
         description: Will be used as part of the cache key. Especially important for unzipped-jdk-source
         type: string
       enrich_classpath:
-        description: Whether the enrich-classpath will be applied.
+        description: Whether enrich-classpath will be applied.
         type: string
       files:
         description: Files to consider when creating the cache key
@@ -160,7 +160,7 @@ jobs:
         description: Maps directly to the TEST_PROFILES var in Makefile
         type: string
       enrich_classpath:
-        description: Whether the enrich-classpath will be applied.
+        description: Whether enrich-classpath will be applied.
         type: string
     executor: << parameters.jdk_version >>
     environment:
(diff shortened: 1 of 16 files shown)


bbatsov:
> Remove constructor info for now, given it's was incorrect and incomplete (and unused from cider.el) for both Reflector and Sources 

What exactly is this constructor info you're referring to? I recall constructors had normal arglists (just with multiple signatures). 

 > Implement and return :formatted-arglists for CIDER usage

Why "formatted"? Something like "annotated" or `typed` arglists will probably be a more fitting name.

bbatsov:
> e.g. for a given constructor with 4 different signatures, maybe just 1 was being shown, so it wasn't reliable enough.

I thought we showed all possible signature together in such instances.

vemv:
> What exactly is this constructor info you're referring to? I recall constructors had normal arglists (just with multiple signatures).

Yes, that's we want is that and it mostly works, but it has enough inconsistencies to prevent a good solution from being built today.

I'd propose, let's keep it disabled for the extent of this PR, and the immediate next PR will re-enable them with all the changes needed for them to be correct and complete.

Otherwise I'd further bloat this PR.

At no point Orchard will be released in this transient state.

> Why "formatted"? Something like "annotated" or typed arglists will probably be a more fitting name.

I initially went for "Formatted" because in the end `^void [this, ^function.BiConsumer action]` is just one particular formatting - using Clojure-style syntax to describe Java arglists is a style/format that while is effective communication, it's rarely used in any context (that I know of).

However `annotated` seems reasonable to me if you're not convinced 👍

bbatsov:
Let's go with `annotated`.

vemv:
Finally green after _many_ little annoyances...

I'll check this once again tomorrow with a fresher mind, then merge

PR #194, opened 2023-10-05
Don't infer indentation for `def`-like macros

vemv:
def-like macros should get no inference - these tend to be risky to make any assumption about.

I actually never particularly meant for def-like macros to be applied these rules of thumb - it was an oversight in implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bugs fixed
 
 * Correctly infer indentation for `->`-like macros.
+* Don't infer indentation for `def`-like macros.
 
 ## 0.15.1 (2023-09-21)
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Hmm, I don't quite understand what's your concern with them.

vemv:
Def-like can mean something like def, or defn, or defprotocol... I'd find it quite adventurous to emit something like `[1 :defn]` out of a guess.

However, if the user chooses exactly the name defn (or def, defprotocol), _and_ the arglists structurally match their clojure.core counterpart, we can confidently infer an indent value identical to that of defn.

So, as originally envisioned, we're encouraging users to choose idiomatic names (defn, not my-defn) and full arglists.

PR #195, opened 2023-10-05
Restore constuctor info

vemv:
Restores all the constructor info that was temporarily disabled in https://github.com/clojure-emacs/orchard/pull/192, expanding related test coverage.


vemv:
I'm going to merge this, since a cider-nrepl release is due, but an async 👀 is welcome.

PR #196, opened 2023-10-19
`orchard.inspect`: offer new `previous-sibling` and `next-sibling` functions

vemv:
> Part of https://github.com/clojure-emacs/cider/issues/3529

I can confirm these work as intended (as I already tried using them in cider / cider-nrepl)

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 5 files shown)


vemv:
Merging as it's a fairly straightforward one - most of the impl resides in the `up`/`down` functions, untouched.

PR #202, opened 2023-10-27
[inspect] Right-align indices when rendering indexed collections

alexander-yakushev:
When inspecting a linear collection, the indices cause an ugly indentation shift on digit promotions (9->10, 99->100). This PR fixes this. Here's now it looks currently:

image

Here's how it looks after the patch:
image

In the future, we might want to do the similar for associative collections, but I think it's a more contentious topic. Some people don't like their values aligned to keys if it causes a large gap on the other side (me included). But aligning indices shouldn't be controversial since most of the time it's gonna be 1 space, 2 spaces at the very most in the edge case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Changes
+
+* [#202](https://github.com/clojure-emacs/orchard/issues/202): `orchard.inspect`: right-align indices when rendering indexed collections.
+
 ### Bugs fixed
 
 * `orchard.inspect`: don't render `Datafy` sections identical to the data they refer to, for nil-valued maps.
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -409,16 +409,22 @@
           mappable))
 
 (defn render-indexed-values
-  ([inspector obj] (render-indexed-values inspector obj 0))
-  ([inspector obj idx-starts-from]
-   (loop [ins inspector, obj (seq obj), idx idx-starts-from]
-     (if obj
-       (recur (-> ins
-                  (render-indent (str idx) ". ")
-                  (render-value (first obj))
-                  (render-ln))
-              (next obj) (inc idx))
-       ins))))
+  "Render an indexed collection of values. Renders all values in `chunk`, so
+  `chunk` must be finite."
+  ([inspector chunk] (render-indexed-values inspector chunk 0))
+  ([inspector chunk idx-starts-from]
+   (let [n (count chunk)
+         last-idx (+ idx-starts-from n -1)
+         last-idx-len (count (str last-idx))
+         idx-fmt (str "%" last-idx-len "s")]
+     (loop [ins inspector, chunk (seq chunk), idx idx-starts-from]
+       (if chunk
+         (recur (-> ins
+                    (render-indent (format idx-fmt idx) ". ")
+                    (render-value (first chunk))
+                    (render-ln))
+                (next chunk) (inc idx))
+         ins)))))
 
 (defn last-page [{:keys [current-page page-size]} obj]
   (cond
--- a/test/orchard/inspect_test.clj
+++ b/test/orchard/inspect_test.clj
@@ -550,7 +550,59 @@
                       '(:newline)
                       "  " "3" ". " (list :value "3" number?)
                       '(:newline))
-                (render (inspect/start (inspect/fresh) [1 2 nil 3]))))))
+                (render (inspect/start (inspect/fresh) [1 2 nil 3])))))
+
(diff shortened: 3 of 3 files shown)


src/orchard/inspect.clj
```diff
@@ -408,17 +408,31 @@
           inspector
           mappable))
 
+(defn- left-pad
+  "TODO: should extract this into a library."
```
vemv, line 412: [oh no!](https://www.theverge.com/2016/3/24/11300840/how-an-irate-developer-briefly-broke-javascript) 😄

Looks like `(String/format "%10s" (object-array ["foo"]))` ensures a given string has N characters, left-padding with spaces otherwise.

src/orchard/inspect.clj
```diff
@@ -408,17 +408,31 @@
...
-                  (render-indent (str idx) ". ")
-                  (render-value (first obj))
-                  (render-ln))
-              (next obj) (inc idx))
-       ins))))
+  "Render an indexed collection of values. Renders all values in `chunk`, so
+  `chunk` must be finite."
+  ([inspector chunk] (render-indexed-values inspector chunk 0))
+  ([inspector chunk idx-starts-from]
+   (let [n (count chunk)
```
vemv, line 425: May it make sense to use bounded-count, when available, for safety?

An altenative being asserting that the chunk is Counted

test/orchard/inspect_test.clj
```diff
@@ -550,7 +550,56 @@
...
+                      '(:newline)
+                      "  " " 5" ". " (list :value "5" number?)
+                      '(:newline)
+                      "  " " 6" ". " (list :value "6" number?)
+                      '(:newline)
+                      "  " " 7" ". " (list :value "7" number?)
+                      '(:newline)
+                      "  " " 8" ". " (list :value "8" number?)
+                      '(:newline)
+                      "  " " 9" ". " (list :value "9" number?)
```
vemv, line 580:
```suggestion
                      "  " " 9" ". " (list :value "9" number?)
                      ;; note how the " 9" has padding and the "10" not, achieving alignment:
```

test/orchard/inspect_test.clj
```diff
@@ -550,7 +550,56 @@
...
+                      '(:newline)
+                      "  " "10" ". " (list :value "10" number?)
+                      '(:newline))
+                (render (inspect/start (inspect/fresh) (vec (range 11))))))
+
+    (let [rendered (-> (inspect/fresh)
```
vemv, line 586: I'd appreciate a `testing` here to understand how this test is different

test/orchard/inspect_test.clj
```diff
@@ -550,7 +550,56 @@
...
+          head (take 11 rendered)
+          tail (take-last 5 rendered)]
+      (is (match? (list "Class"
+                        ": " (list :value "clojure.lang.PersistentVector" number?)
+                        '(:newline)
+                        '(:newline)
+                        "--- Contents:"
+                        '(:newline)
+                        "  " "  0" ". " (list :value "0" number?))
+                  head))
```
vemv, line 599:
```suggestion
                  head))
                  ;; note how the "  0" has padding and the "100" not, achieving alignment:
```

vemv (commented):
Looking good! Happy to get the Ins closer to pixel perfection

src/orchard/inspect.clj
alexander-yakushev: Interesting, I've tried googling for it and found answers that String.format can only pad with zeros.

Still, I feel a bit reluctant to construct a format string and then use the slow formatting method for each iteration, seems a bit wasteful. Even though it will never be used for that many items anyway.

Yeah, it's probably nothing.

src/orchard/inspect.clj
alexander-yakushev: The current version of this function goes over the whole sequence non-conditionally anyway. Adding `count` doesn't really make it less safe. There is currently only one caller of this function that passes in a finite sequence. I renamed the argument to make this more apparent at least documentation-wise. Don't know if it needs any other guardrails.

alexander-yakushev:
Ready.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#202](https://github.com/clojure-emacs/orchard/issues/202): `orchard.inspect`: right-align indices when rendering indexed collections.
```
vemv, line 5:
```suggestion
### Changes

* [#202](https://github.com/clojure-emacs/orchard/issues/202): `orchard.inspect`: right-align indices when rendering indexed collections.
```

vemv (approved):
Feel free to merge after it's green

PR #205, opened 2023-10-29
Make `orchard.java.parser-utils/parse-java-test` more easily debuggable

vemv:
We used to return nil for invalid .java source code.

Now we throw an exception, with the parsing error diagnostics as part of the ex-data.

Note that this doesn't change user-facing behavior, since all exceptions are caught by default in `orchard.java.parser/source-info` / `orchard.java.parser-next/source-info`.

This helps debug faulty .java source code, see e.g. https://github.com/aws/aws-sdk-java/issues/3045

Cheers - V


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changes
 
 * [#202](https://github.com/clojure-emacs/orchard/issues/202): `orchard.inspect`: right-align indices when rendering indexed collections.
+* Internal: make `orchard.java.parser-utils/parse-java-test` more easily debuggable.
```
bbatsov, line 8: Such changes don't need a changelog entry. (I guess end users are not supposed to use this function directly)

bbatsov:
I'm fine with the proposed change.

CHANGELOG.md
bbatsov: I normally never document internal changes unless they can be observed externally somehow (e.g. performance improvement). I think that's just noise for many of the users. At any rate - I don't feel strongly about this, so feel free to keep it.

vemv: 👍 Removing

PR #208, opened 2023-11-03
Introduce new `read-ns-form`, `project-ns-forms`, `ns-form->imports` functions

vemv:
Supports https://github.com/clojure-emacs/cider-nrepl/issues/830

The idea is to provide a means to easily get a list of `:import`ed classes, even if the underlying namespaces haven't been loaded yet.

Cheers - V

### DIFF
--- a/src/orchard/namespace.clj
+++ b/src/orchard/namespace.clj
@@ -16,9 +16,16 @@
 
 ;;; Namespace/source resolution
 
-(defn read-namespace
-  "Returns the namespace name from the first top-level `ns` form in the file."
-  [url]
+(defn- ns-form->ns-name [x]
+  (let [s (second x)]
+    (when (symbol? s)
+      s)))
+
+(defn- ns-form? [x]
+  (and (list? x)
+       (-> x first #{`ns 'ns})))
+
+(defn- read-ns [url pred extract]
   (with-open [r (PushbackReader. (io/reader url))]
     (loop []
       (let [found (try
@@ -29,12 +36,33 @@
                     (catch Exception _
                       ::fail))]
         (cond
-          (#{::eof ::fail} found) nil
-          (and (list? found)
-               (-> found first #{`ns 'ns})
-               (symbol? (second found)))
-          (second found)
-          :else (recur))))))
+          (#{::eof ::fail} found)
+          nil
+
+          (pred found)
+          (extract found)
+
+          :else
+          (recur))))))
+
+(defn read-ns-name
+  "Returns the namespace name from the first top-level `ns` form in the file."
+  [url]
+  (read-ns url
+           (every-pred ns-form? ns-form->ns-name)
+           ns-form->ns-name))
+
+(defn read-namespace
+  "Returns the namespace name from the first top-level `ns` form in the file."
+  {:deprecated "0.19"}
+  [url]
+  (read-ns-name url))
+
+(defn read-ns-form
+  "Returns the first top-level `ns` form in the file."
+  {:added "0.19"}
+  [url]
+  (read-ns url ns-form? identity))
 
 (defn canonical-source
   "Returns the URL of the source file for the namespace object or symbol,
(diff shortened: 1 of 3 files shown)


src/orchard/namespace.clj
```diff
@@ -135,6 +167,46 @@
        (filter identity)
        (classpath-namespaces)))
 
+(defn project-jvm-ns-forms
```
bbatsov, line 170: Isn't much of this ns JVM specific anyways? Seems weird to use a different naming for just a single function.

src/orchard/namespace.clj
```diff
@@ -29,12 +36,26 @@
...
+          (#{::eof ::fail} found)
+          nil
+
+          (pred found)
+          (extract found)
+
+          :else
+          (recur))))))
+
+(defn read-namespace
```
bbatsov, line 48: Perhaps `ns-name` instead? `read-namespace` could also be interpreted as `ns-form`. (the function below)

src/orchard/namespace.clj
```diff
@@ -104,8 +125,16 @@
 ;; These methods search sources on the classpath. Non-classpath source
 ;; files, documentation code, etc within the project directory are ignored.
 
+(defn jvm-clojure-resource-name->ns-form
```
bbatsov, line 128: I'm not sure that `jvm` here is needed, given how much of the codebase in Orchard is (Java) Clojure-specific anyways.

src/orchard/namespace.clj
```diff
@@ -135,6 +167,46 @@
...
+                               (when ((every-pred misc/directory? in-project?) x)
+                                 x)))
+                       (filter identity))]
+    (into {}
+          (map (fn [ns-form]
+                 [(ns-form->ns-name ns-form)
+                  ns-form]))
+          (classpath-namespaces resources jvm-clojure-resource-name->ns-form))))
+
+(defn ns-form->imports
```
bbatsov, line 185: That's not really a conversion function, so I'd probably name it something like `ns-imports` or `get/extract-ns-imports`.

bbatsov:
Looks good to me overall, apart from my small remarks. It'd be good to add `:added` metadata to the new public functions, so it's clear when they were introduced.

src/orchard/namespace.clj
vemv: It's its existing name, but we can deprecate it in favor of that name?

src/orchard/namespace.clj
vemv: I'd hope that to be decreasingly the case over time!

src/orchard/namespace.clj
vemv: IMO `project-namespaces` should have been named `project-jvm-namespaces`. In the end it's just a tweak away from being cljs-compatible.

I'd propose to deprecate `project-namespaces` in favor of the alternative, laying out the path forward.

src/orchard/namespace.clj
vemv: You might want to expand on what is / is not a conversion function here https://github.com/bbatsov/clojure-style-guide#naming-conversion-functions

src/orchard/namespace.clj
bbatsov: What an existing name? `read-ns-name` or `ns-name`?

src/orchard/namespace.clj
bbatsov: Or it can just a be a `orchard.java.namespace` (similar to some namespaces in Clojure and other projects). I don't like including implementation details in the names of functions and introducing internal naming inconsistencies. I understand your point, but given there's no ongoing push to add ClojureScript support this seems like a case of premature optimization.

src/orchard/namespace.clj
bbatsov: Ah, `read-namespace` was moved. Yeah, I guess it can be renamed and the old name can be deprecated.

src/orchard/namespace.clj
bbatsov: Yeah, but as noted elsewhere I think it'd be weird to prefix all platform-specific functions here. Probably it'd be better to have implementation specific namespaces and some platform-independent ns that wraps around those.

vemv: Note that there was already an existing `jvm-clojure-resource-name->ns-name` defn, so it's consistent.

Similarly, I renamed `project-jvm-ns-forms` -> `project-ns-forms` for consistency with what's here

src/orchard/namespace.clj
vemv: For now I'll leave things are they were before, even if they weren't ideal.

We can refactor later indeed, driven by a specific need.

src/orchard/namespace.clj
vemv: Done!

src/orchard/namespace.clj
vemv: Renamed

Issue #211, opened 2023-11-11
Look into Roaster


vemv:
https://github.com/forge/roaster is a Java parser that appears to be much faster than JDK's (1-20ms vs. N seconds).

It's based on [Eclipse JDT](https://projects.eclipse.org/projects/eclipse.jdt).

It also seems less picky e.g. it can parse [Lomboked](https://github.com/clojure-emacs/enrich-classpath/issues/48#issuecomment-1806687843) classes.

Sample usage was easy:

    (time (do (Roaster/parse (io/resource (orchard.java.parser-utils/source-path `Thread))) nil))

alexander-yakushev:
After performance optimizations done to Orchard, and by removing aggressive pre-caching, the native Java parser has been working sufficiently fast for this no longer being an issue. Closing for now.

PR #212, opened 2023-11-11
`orchard.java`: add `*analyze-sources*` dynamic var

vemv:
Haystack's Java needs are rather humble https://github.com/clojure-emacs/haystack/blob/3b0d8a464f60cb0c0e4114b9d853165d6d2f59ae/src/haystack/analyzer.clj#L91-L95 and we were responding slowly to those requests as we were parsing the underlying .java files when possible.

The introduced dyn var makes it possible for consumers to request less info, faster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
+## New features
+
+* `orchard.java`: add `*analyze-sources*` dynamic var.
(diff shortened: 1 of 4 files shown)


vemv:
I'll merge this right away as I consider fixing this quite urgent.

Mid-term, we should look into an [alternative parser](https://github.com/clojure-emacs/orchard/issues/211).

PR #213, opened 2023-11-12
Introduce `orchard.inspect/nav-to-item`

### LINKED ISSUE #209: Inspector: add op to nav a sub item
Unlike e.g. Portal, we don't have a command to `nav` to a sub-item (e.g. nth, or k-v entry) and re-render the Inspector using the produced value.

vemv:
Closes https://github.com/clojure-emacs/orchard/issues/209

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -120,11 +120,14 @@
         (assoc :path new-path)
         (inspect-render new))))
 
+(defn- nth-wrapper? [x]
+  (and (seq? x)
+       (= 'nth (first x))))
+
 (defn- sibling* [inspector offset pred]
   (let [path (:path inspector)
         last-item (peek path)]
-    (or (when (and (seq? last-item)
-                   (= 'nth (first last-item)))
+    (or (when (nth-wrapper? last-item)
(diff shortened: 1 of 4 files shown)


src/orchard/inspect.clj
```diff
@@ -508,19 +539,25 @@
         (unindent))
     inspector))
 
-(defn- nav-datafy-tx [obj remove-nil-valued-entries?]
+(defn- nav-datafy-tx [remove-nil-valued-entries?]
   (keep (fn [[k v]]
-          (or (some->> (nav obj k v)
```
vemv, line 513: Note the deletion of `nav`, explained in the comment below.

I'd say it was a bug - deftests were showing that we were performing `nav` over the same items repeatedly, which can be expensive in the face of side-effects.

test/orchard/inspect_test.clj
```diff
@@ -941,32 +958,6 @@
                         (:newline))
                       (datafy-section rendered))))))))
 
-(deftest inspect-navigable-metadata-extension-test
```
vemv, line 944: Test deleted as it was based on eagerly calling `nav`.

The added test already shows the behavior for when calling nav intentionally.

vemv:
Hmm, the basic idea works but this will need some refinements since the `index` has some edge cases.

Issue #217, opened 2023-11-14
Inspector: Classes lack a header


vemv:
If I'm inspecting the class `clojure.lang.PersistentHashMap`, the very first thing one should see rendered should be something like `Name: clojure.lang.PersistentHashMap`.

<img width="757" alt="image" src="https://github.com/clojure-emacs/orchard/assets/1162994/62483012-78e1-42fc-8f5c-ccf1a6ce1eff">

This is a title that gives users context (otherwise somewhat easy to forget)

PR #221, opened 2023-12-04
Add a `Name` header to inspected Class objects

vemv:
> Closes https://github.com/clojure-emacs/orchard/issues/217

Adds a `Name` header to inspected Class objects.

Also makes some tests less fragile - they were breaking on JDK 21.

Cheers - V

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -44,7 +44,10 @@
     (conj path '<unknown>)
     (if (= idx 0)
       (conj path 'class)
-      (let [klass (first index)]
+      (let [klass (first index)
+            klass (if (class? klass)
+                    klass
+                    (-> klass str Class/forName))]
         (cond
           ;; If value's class is a map, going down means jumping into either key
           ;; or value.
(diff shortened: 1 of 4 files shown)


src/orchard/inspect.clj
```diff
@@ -686,7 +689,7 @@
 
 (defn- render-section [obj inspector [section sort-key-fn]]
   (let [method (symbol (str ".get" (name section)))
-        elements (eval (list method obj))]
+        elements (eval (list method obj))] ;; XXX ditch eval
```
bbatsov, line 692: XXX? Should be `TODO:` or `FIXME:`?

bbatsov:
I'm kind of puzzled what exactly is the difference between "Name" and "Class", though, as it seem the name is derived from the class.

vemv:
If we're inspecting the class object `java.lang.String`, its name is `java.lang.String` and its class is `java.lang.Class`

bbatsov:
But inspecting classes will be the only case in which those 2 are different, right? I'm just wondering if "Name" is the right name for this.

vemv:
Yes, only `defmethod inspect :class` is affected

src/orchard/inspect.clj
vemv: > XXX? Should be TODO: or FIXME:?

fixed by actually addressing the thing 😄

PR #223, opened 2024-01-01
Optimize the performance of namespace discovery

### LINKED ISSUE #222: Use dir globbing to discover .clj files
`orchard.namespace` can iterate over potentially many, many files (200K+ for a specific user) to determine whether they're .clj files. That's slow enough to cause nrepl timeouts.

Using dir globbing would be faster https://stackoverflow.com/questions/9148528/how-do-i-use-directory-globbing-in-jdk7

See also https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-

vemv:
Fixes #222

Will confirm whether it's an actual improvement soon :)

### DIFF
--- a/src/orchard/namespace.clj
+++ b/src/orchard/namespace.clj
@@ -154,8 +154,11 @@
 (defn- find-cljc-files-efficiently
   "Finds .clj/c files as efficiently as possible.
 
-  In particular, avoids visiting and operating on files,
-  as early as possible."
+  In particular, avoids returning unnecessary files,
+  and creating objects for those.
+
+  This way, upstream consumers like `orchard.java.classpath/classpath-seq` will operate on fewer File objects,
+  improving performance for large workloads (e.g. a directory with 1M files was placed as a resource)."
   [^File dir]
   (let [start-path (.toPath dir)
         matcher (reify BiPredicate
--- a/src/orchard/namespace.clj
+++ b/src/orchard/namespace.clj
@@ -11,8 +11,11 @@
    [orchard.java.classpath :as cp]
    [orchard.misc :as misc])
   (:import
-   (java.io PushbackReader)
-   (java.net URL)))
+   (java.io File PushbackReader)
+   (java.net URL)
+   (java.nio.file FileVisitOption Files Path)
+   (java.util.function BiPredicate)
+   (java.util.stream Collectors)))
 
 ;;; Namespace/source resolution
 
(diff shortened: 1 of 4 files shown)


vemv:
@bbatsov ready, sorry for the excess commits.

The affected user could test out this branch and confirms a much better performance for his scenario (1M files in a dir in the classpath)

bbatsov:
How much better? 

I understand the problem, but I'm wondering whether the complexity trade-off is worth it as it seems to me that's a pretty rare case.

vemv:
Thanks for the review!

The improvements is from "Emacs freezes / nrepl times out" to 1000ms (only experienced when we use `(project-namespaces)` e.g. when we analyze stacktraces)

Besides from that improvement, I value that this otherwise hard to debug problem mostly disappears. It wasn't trivial to diagnose this.

At first I feared that this was a "Java parser" type of problem (those plagued stacktrace analysis some months ago). Ruling that out is useful to me.

src/orchard/namespace.clj
```diff
@@ -148,9 +151,27 @@
             io/resource ;; can return nil for Emacs backup files, for example
             read-ns-name)))
 
+(defn- find-cljc-files-efficiently
+  "Finds .clj/c files as efficiently as possible.
+
+  In particular, avoids visiting and operating on files,
```
bbatsov, line 157: I'm struggling to understand what this means exactly. :-)

src/orchard/namespace.clj
```diff
@@ -148,9 +151,27 @@
...
+  as early as possible."
+  [^File dir]
+  (let [start-path (.toPath dir)
```
bbatsov, line 160: I'd suggest to document in more details the implementation details here, as I'm not familiar with neither `BiPredicate` and the related Java APIs, so it's hard for me to comment on the quality of the implementation. But I'd definitely like to understand better why are we using those.

vemv: I (hopefully) improved the docstring.

The improvement gains are obtained as follows:

* We have an optimized 2-arg predicate (BiPredicate) that shortcircuits / avoids garbage creation
* We return fewer files (only .clj/c files)
* Because it receives fewer files, the upstream consumer (`orchard.java.classpath/classpath-seq`) doesn't uselessly create more computations/garbage e.g.:

https://github.com/clojure-emacs/orchard/blob/951fdbe26cf5146ef070b905a99508e6ef9997e1/src/orchard/java/classpath.clj#L91-L93

For the affected user's resource dir, we now run the code above for 50 .clj/c files instead of 1M irrelevant files.

Other than that, there's no particular magic in the Java interop, it's a fairly vanilla way of listing files (vs. `file-seq` which lacks any filtering options).

If we wanted to further optimize things, we could look into performing a Unix `ls | grep` with C interop, which I believe would be faster since it avoids creating 1M `Path`/`File`objects.

But things seem usable as-is at the moment.

bbatsov: Yeah, the interop code is not particularly complex - I was mostly wondering if we need it at all and can get similar results with Clojure's file APIs.

src/orchard/namespace.clj
```diff
@@ -148,9 +151,30 @@
             io/resource ;; can return nil for Emacs backup files, for example
             read-ns-name)))
 
+(defn- find-cljc-files-efficiently
```
bbatsov, line 154: I'd probably use a different name for this, as it now seems it looks for "cljc" files. Perhaps you can make the extensions params, so this can be a more generic function instead?

src/orchard/namespace.clj
bbatsov: The docstring is fine in its current form.

src/orchard/namespace.clj
vemv: We could use `find-clj+cljc-files-efficiently`?

The param is an acceptable option, OTOH making things generic would mean that we can't longer accurarately use the docstring for describing what we're gaining.

bbatsov: I'd be OK with the name, although I don't quite understand what would be the issue with the docstring if the function was generic, though - it would be similar to the current one, but with a note that you have to specify the file extensions you're targetting.

PR #224, opened 2024-01-15
Catch NoClassDefFoundError

hugoduncan:
This exception can be thrown on a case indifferent filesystem, eg. on
macOS.

One way this happens is if there is a java class with a camel cased name
that otherwise matches the name of a clojure namespace.

For some reason, when this exception was thrown, the classes in the
class cache were also getting initialised.

### DIFF
--- a/src/orchard/meta.clj
+++ b/src/orchard/meta.clj
@@ -103,9 +103,9 @@
          (catch ClassNotFoundException _
            nil)
          (catch java.lang.NoClassDefFoundError _
-           ;; this can happen if there is a java class with a camel cased name
+           ;; This can happen if there is a java class with a camel cased name
            ;; that otherwise matches the name of a clojure namespace, on a case
-           ;; indifferent filesystem, eg. on OS X.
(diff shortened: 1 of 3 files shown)


vemv:
Looking reasonable to me - thanks!

src/orchard/meta.clj
```diff
@@ -102,6 +102,11 @@
          ;; Impl might try to resolve it as a class, which may fail
          (catch ClassNotFoundException _
            nil)
+         (catch java.lang.NoClassDefFoundError _
+           ;; this can happen if there is a java class with a camel cased name
+           ;; that otherwise matches the name of a clojure namespace, on a case
+           ;; indifferent filesystem, eg. on OS X.
```
bbatsov, line 108: I believe it's named macOS these days. :-)

bbatsov:
Same, just needs a changelog entry.

vemv:
Thanks! Would you need a release asap? Otherwise I'd bundle it with other stuff next weekend

hugoduncan:
No rush, thanks.

PR #225, opened 2024-01-15
[ci] Refactor CI wrt JDK8 and Clojure 1.9 compatibility

alexander-yakushev:
The goal is to make Orchard CI config match the configs of cider-nrepl (https://github.com/clojure-emacs/cider-nrepl/pull/841) and nREPL (https://github.com/nrepl/nrepl/pull/305).


bbatsov:
The proposed changes look good to me. Thanks!

PR #226, opened 2024-02-17
`orchard.xref`: avoid occasional null pointer access

vemv:
Performing `xref` over an invalid symbol (e.g. not a function) could result in an `eval` that would not return a `Class` object, which would then make the `.getDeclaredFields` call invalid.

### DIFF
--- a/test/orchard/xref_test.clj
+++ b/test/orchard/xref_test.clj
@@ -20,7 +20,7 @@
   (is (nil? (xref/fn-deps-class 2))
       "Is garbage-safe (important, as it uses `eval` which can return anything)")
 
-  (is (set/superset? (xref/fn-deps-class (.getClass xref/fn-deps-class))
+  (is (set/superset? (xref/fn-deps-class (.getClass ^Object xref/fn-deps-class))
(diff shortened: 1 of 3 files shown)


bbatsov:
Shouldn't we add some unit test for this?

vemv:
Thought so but we hadn't any for the defn to begin with. Can do 👍

vemv:
Ready

test/orchard/xref_test.clj
```diff
@@ -70,6 +71,9 @@
 (def yyy (symbol (str (gensym))
                  (str (gensym))))
 
+(not (= #{#'orchard.xref-test/fn-deps-test #'clojure.core/filter #'orchard.xref-test/fn-transitive-dep #'clojure.core/inc' #'clojure.core/range #'clojure.test/test-var #'clojure.core/even? #'clojure.core/map #'orchard.xref-test/fn-dep}
```
bbatsov, line 74: What's the purpose of this code?

vemv: Huh, it slipped by

PR #228, opened 2024-03-02
`info*`, offer `:var-meta-allowlist` option

vemv:
> Part of https://github.com/clojure-emacs/cider-nrepl/issues/851

It allows consumers to specify var metadata beyond Orchard's fixed whitelist.

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+## 0.23.0 (2024-03-03)
+
 ## Changes
 
 * [cider-nrepl#851](https://github.com/clojure-emacs/cider-nrepl/issues/851): `info*`, offer `:var-meta-allowlist` option.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## master (unreleased)
 
+## Changes
+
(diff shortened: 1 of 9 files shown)


vemv:
@bbatsov PR ready, the `master` builds are red due to Clojure 1.12

(investigating this weekend)

vemv:
The fix was actually easy - done.

bbatsov:
These days I try to avoid the whitelist terminology for various reasons, so let's name this something else - e.g. `var-meta-keys` or something along those lines. (basically something reflecting better we are selecting only a subset of the var metadata - meaningful keys, useful keys, etc) Let's also add an alias to the old var matching whatever name we pick here.

vemv:
Good catch! Will refactor

vemv:
Ready

bbatsov:
I meant using something more specific than a generic name like `allowlist`, but anyways - that's not a big deal. The conversation around `blacklist` vs `allowlist` also made me realize how bad such names are in general, as they don't carry much meaning. In our case we want only a certain properties because we don't care about the others and they might have content that we can't parse. That's not really an allowlist, but something more specific IMO, but I guess I didn't manage to convey this clearly.

vemv:
Ah, I see, sorry about that. Could have been `selected-keys` (following `select-keys`).

Although there's some value in using a term that is well-understood, even if it doesn't always exactly map to what it means.

PR #229, opened 2024-03-04
Minor changes

p4v4n:


test/orchard/inspect_test.clj
```diff
@@ -447,9 +447,9 @@
 
 (deftest path-test
   (testing "inspector tracks the path in the data structure"
```
p4v4n, line 449: This test was frequently failing on a local wip branch after some changes to inspect/down fn. Changing it to a sorted-map to make debugging easier.

Issue #216, opened 2023-11-14
Inspector next-sibling doesn't work past page 1


vemv:
Looks like the next-sibling fn is buggy past page 1.

PR #230, opened 2024-03-05
Fix inspector navigation fns to work with pagination

p4v4n:
Closes https://github.com/clojure-emacs/orchard/issues/216

- Makes `next-sibling`, `previous-sibling`, and `down` fns work well with pagination.
- `next-sibling` will continue to not go beyond the current page.
- Adds out-of-bound checks for `down`, `next-sibling`, `next-page`, and `previous-page` fns.
- Handle discrepancy with page-size and items-on-page when inspecting hash-maps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs Fixed
+
+* [#216](https://github.com/clojure-emacs/orchard/issues/216): Fix inspector navigation (`down` and `next-sibling`) with pagination.
+
 ## 0.23.0 (2024-03-03)
 
 ## Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Bugs Fixed
 
 * [#216](https://github.com/clojure-emacs/orchard/issues/216): Fix some pagination issues with inspector navigation.
-  * `down` fn goes beyond current page. `next-sibling` and `previous-sibling` fns work in all pages.
+  * `orchard.inspect/down` fn is now able to go beyond the current page. `next-sibling` and `previous-sibling` fns work in all pages.
 
 ## 0.23.0 (2024-03-03)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Bugs Fixed
 
-* [#216](https://github.com/clojure-emacs/orchard/issues/216): Fix inspector navigation (`down` and `next-sibling`) with pagination.
+* [#216](https://github.com/clojure-emacs/orchard/issues/216): Fix some pagination issues with inspector navigation.
+  * `down` fn goes beyond current page. `next-sibling` and `previous-sibling` fns work in all pages.
 
 ## 0.23.0 (2024-03-03)
 
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -144,14 +144,18 @@
 (defn next-page
   "Jump to the next page when inspecting a paginated sequence/map. Does nothing
   if already on the last page."
-  [inspector]
-  (inspect-render (update-in inspector [:current-page] inc)))
+  [{:keys [current-page] :as inspector}]
+  (if (= current-page (last-page inspector))
+    inspector
+    (inspect-render (update inspector :current-page inc))))
 
 (defn prev-page
   "Jump to the previous page when inspecting a paginated sequence/map. Does
   nothing if already on the first page."
-  [inspector]
-  (inspect-render (update-in inspector [:current-page] dec)))
+  [{:keys [current-page] :as inspector}]
+  (if (zero? current-page)
+    inspector
+    (inspect-render (update inspector :current-page dec))))
 
 (defn up
   "Pop the stack and re-render an earlier value."
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -117,9 +117,6 @@
   ([inspector] (last-page inspector (:value inspector)))
   ([{:keys [current-page page-size]} obj]
    (cond
-     (map? obj)
-     (quot (dec (* 2 (count obj))) page-size)
-
      (instance? clojure.lang.Counted obj)
      (quot (dec (count obj)) page-size)
 
@@ -145,7 +142,7 @@
   "Jump to the next page when inspecting a paginated sequence/map. Does nothing
   if already on the last page."
   [{:keys [current-page] :as inspector}]
-  (if (= current-page (last-page inspector))
+  (if (>= current-page (last-page inspector))
     inspector
     (inspect-render (update inspector :current-page inc))))
 
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -95,6 +95,13 @@
   (-> (clear inspector)
       (inspect-render value)))
 
+(defn- items-on-page
+  "Number of individual top level items on a single page."
+  [{:keys [value page-size] :as _inspector}]
+  (if (map? value)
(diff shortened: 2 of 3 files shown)


src/orchard/inspect.clj
```diff
@@ -95,6 +95,18 @@
   (-> (clear inspector)
       (inspect-render value)))
 
+(defn next-page
+  "Jump to the next page when inspecting a paginated sequence/map. Does nothing
+  if already on the last page."
+  [inspector]
+  (inspect-render (update-in inspector [:current-page] inc)))
```
vemv, line 102: Shouldn't be there some sort of checking before `inc`?

I believe that's the root of our issues.

vemv (commented):
Very happy that you're taking on this one!

Some heads up

src/orchard/inspect.clj
p4v4n: Is the check for not going beyond the last page?
That will be a good addition too but it's not related to the current issue IMO.

p4v4n: The main issues I fixed here are
1) down fn was not going beyond the first page.
2) Handle the discrepancy between page-size and no: of items on the page when a hash-map is being inspected. This discrepancy IIUC existed since the separation of key/values was implemented [here](https://github.com/clojure-emacs/orchard/pull/24).

bbatsov:
@alexander-yakushev feel free to take a look as well.

vemv (commented):
Looking reasonable

test/orchard/inspect_test.clj
```diff
@@ -381,6 +381,37 @@
...
+    (is (= 9 (-> long-map
+                 inspect
+                 (inspect/set-page-size 2)
+                 (inspect/down 20)
+                 :value)))
+    (is (= 19 (-> long-map
+                  inspect
+                  (inspect/down 40)
+                  :value)))))
+
```
vemv, line 414: Note that there's `(deftest sibling*-test` - shouldn't it be extended?

vemv (commented):
Looks like GH ate my comment - here it goes

src/orchard/inspect.clj
p4v4n: I have fixed the new issues at the cost of removing the safety check in next-sibling to not go beyond the last element. 
I am not sure if it's possible to know the total size of the inspected paginated collection. 
This check might have to be removed anyways when supporting next-sibling for other collection types.

vemv: Hmm, ideally we'd only merge things that improve some aspect without regressing on others.

Perhaps you can decompose this PR into the bits that you are most sure about?

(Admittedly some of these problems are pretty involved, which is why I left them pending)

p4v4n: Not really in a hurry to get this merged. It's probably better to leave it here for a couple of days and think about it more.

I was trying to support next-sibling across pages as well which might not be possible with the same safety check without knowing the full collection size.(1ex would be lazy sequences)

Another fallback option is to support next-sibling within a page (with the same safety check) and use next-page to navigate across pages.

test/orchard/inspect_test.clj
```diff
@@ -409,7 +440,7 @@
              (inspect/previous-sibling)
              (inspect/previous-sibling)
              :value)))
-  (is (= :c
+  (is (= nil
```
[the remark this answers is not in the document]
p4v4n: This is the result of removing the safety check in next-sibling as explained in the other comment.
I will try to add back the safety check without causing other issues.

src/orchard/inspect.clj
vemv: Thanks for understanding and for the sustained efforts - appreciated!

If it makes things simpler, it would be acceptable to not support `next-sibling` on lazy collections at all. They're relatively infrequent (vs vectors, hashmaps) so I'd favor having something correct+complete over the most frequent use cases.

(an elegant way to not support something here is to simply re-render the same Inspector state)

test/orchard/inspect_test.clj
```diff
@@ -381,6 +381,37 @@
         (inspect/def-current-value *ns* "--test-val--"))
     (is (= 1 @(resolve '--test-val--)))))
 
+(deftest down-test
+  (testing "basic down"
+    (is (= :a (-> '{:foo [:a :b :c]}
+                  inspect
+                  (inspect/down 2)
+                  (inspect/down 1)
```
vemv, line 389: Suggested extra coverage: try to go `down` out of boundaries, e.g. `inspect/down 99` in this line and see what happens

(we do that in `sibling*-test`)

Same for pagination-related tests

test/orchard/inspect_test.clj
```diff
@@ -443,13 +474,58 @@
...
+                  (inspect/previous-sibling)
+                  (inspect/previous-sibling)
+                  :value)))
+    (is (= 40 (-> long-vector
+                  inspect
+                  (inspect/down 40)
+                  (inspect/next-sibling)
+                  :value))))
+  (testing "next sibling doesn't go beyond the current page"
+    (is (= 41 (-> long-vector
```
vemv, line 504: Besides from this example, which is OK, please add a small, self-contained example where one can just _see_ the intended behavior.

In other words, as a maintainer it's pretty hard to build a mental model of a `(vec (range 70))` and make reasoning about it (e.g. "41 is good").

As an example that I'd find good, https://github.com/clojure-emacs/orchard/blob/11e191b3f405e7e2a2d6e4a44b42ef26499e58d8/test/orchard/inspect_test.clj#L412-L424 is a small data structure, so if I see seven `(inspect/next-sibling)` calls, it's clear to casual readers that it's trying to go beyond bounds.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## master (unreleased)
 
+### Bugs Fixed
+
+* [#216](https://github.com/clojure-emacs/orchard/issues/216): Fix some pagination issues with inspector navigation.
+  * `down` fn goes beyond current page. `next-sibling` and `previous-sibling` fns work in all pages.
```
vemv, line 8:
```suggestion
  * `orchard.inspect/down` is now able to go beyond the current page. `next-sibling` and `previous-sibling` fns work in all pages.
```

vemv (commented):
Looking much better now - kudos!

Feel free to mark the PR as ready after this round.

test/orchard/inspect_test.clj
p4v4n: Made the other 2 changes.
This one won't produce any predictable value as `down` fn doesn't have an out-of-boundary check. (It might be okay for now as a user cannot trigger this fn directly)

p4v4n: `next-page` fn also has the same problem. Maybe this is what you meant with the first comment in this PR.

I will think about this a bit more and see if it's possible to add checks for both these fns.

vemv: Thanks! Much appreciated

p4v4n: Added out-of-bound checks for `down` and `next-page` fns and tried to cover all the scenarios in tests.

p4v4n: Marking this as ready for review again. Feel free to suggest any more changes/tests/nits.

vemv:
Great work! It's been a pleasure to see coverage grow commit after commit.

I'll try this out locally. Feel free to, as well.

Cheers - V

vemv:
Merging and releasing this in cider-nrepl today

tbh I didn't have the chance to try it out, but based on code review and the test coverage, it certainly looks like a great overall improvement.

I have a little personal app where I use `next-sibling` all the time so any possible improvement won't escape our radar.

Cheers - V

PR #235, opened 2024-03-25
Make orchard.java/class-info* not initialize classes

jpmonettas:
Haven't checked all PR requirements and there are tests failing. Adding this after the slack discussion [here](https://clojurians.slack.com/archives/C0617A8PQ/p1711224784839239) 
This is just for reference. Can be deleted after.

### DIFF
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -323,10 +323,11 @@
   first by name, and then by argument types to list all overloads."
   [class]
   (when-let [^Class c (try
-                        (Class/forName (str class)) ;; NOTE: we don't pass the `false` argumnt since that complicates the analysis for deftype/defrecord classes.
-                        (catch Exception _)
-                        (catch NoClassDefFoundError _)
-                        (catch LinkageError _))]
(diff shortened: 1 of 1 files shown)


vemv:
I'm not sure a PR is the right way to re-surface this topic, given that we observed that there wasn't a trivial fix.

Anyway, I'll ponder on the topic - feel free to ping me if there are no news.

jpmonettas:
@vemv yeah I think this is going to be tricky to solve reliable. But I also think it is important since currently you can't work with Cider on a JavaFX project for example.

Maybe the easiest thing to do is to add some kind of blacklist for skipping initializing classes with a JVM property or something? 
I think this will be a low risk change and provide a solution for people working on this kind of projects and wanting to upgrade Cider (like me :P)

vemv:
A blacklist would SGTM. It can be named e.g. `orchard.edn` with a `:class-info-banlist` k-v.

You can take a look at https://github.com/clojure-emacs/cider-nrepl/pull/863/files as an example of obtaining multiple identically-named resources and then combining them safely.

(Can also do myself)

PR #237, opened 2024-04-11
[java] Improve Java parsing performance

alexander-yakushev:
I've noticed that most of the time spent during running tests is calling `source-info` from different places.

### DIFF
--- a/src-newer-jdks/orchard/java/parser.clj
+++ b/src-newer-jdks/orchard/java/parser.clj
@@ -230,14 +230,14 @@
         (when-let [^DocletEnvironment root (parse-java path (module-name klass))]
           (try
             (let [path-resource (io/resource path)]
-              (assoc (->> (.getIncludedElements root)
-                          (filter #(#{ElementKind/CLASS
-                                      ElementKind/INTERFACE
-                                      ElementKind/ENUM}
-                                    (.getKind ^Element %)))
-                          (keep #(parse-info % root))
-                          (filter #(= klass (:class %)))
-                          (first))
+              (assoc (some #(when (#{ElementKind/CLASS
+                                     ElementKind/INTERFACE
+                                     ElementKind/ENUM}
+                                   (.getKind ^Element %))
+                              (let [info (parse-info % root)]
+                                (when (= (:class info) klass)
+                                  info)))
+                           (.getIncludedElements root))
                      ;; relative path on the classpath
                      :file path
                      ;; Legacy key. Please do not remove - we don't do breaking changes!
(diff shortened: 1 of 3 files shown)


alexander-yakushev:
It doesn't seem to outright help with times on CI, but it does on my machine. I think these changes are useful nonetheless.

alexander-yakushev:
@vemv Is there an explicit reason why Eastwood runs in each test job and not together with Kondo and others in a dedicated job?

src-newer-jdks/orchard/java/parser_next.clj
```diff
@@ -323,14 +323,14 @@
...
           (try
             (let [path-resource (io/resource path)]
-              (assoc (->> (.getIncludedElements root)
-                          (filterv #(#{ElementKind/CLASS
-                                       ElementKind/INTERFACE
-                                       ElementKind/ENUM}
-                                     (.getKind ^Element %)))
-                          (mapv #(parse-info % root))
-                          (filterv #(= klass (:class %)))
-                          (first))
```
alexander-yakushev, line 333: Because of eagerness, this code invoked `parse-info` on every element of `root` even though we only need the first one that matches `(= klass (:class %)`.

src-newer-jdks/orchard/java/parser.clj
```diff
@@ -230,14 +230,14 @@
...
           (try
             (let [path-resource (io/resource path)]
-              (assoc (->> (.getIncludedElements root)
-                          (filter #(#{ElementKind/CLASS
-                                      ElementKind/INTERFACE
-                                      ElementKind/ENUM}
-                                    (.getKind ^Element %)))
-                          (keep #(parse-info % root))
-                          (filter #(= klass (:class %)))
-                          (first))
```
alexander-yakushev, line 240: This is a bit better due to laziness, but it will still realize at least a 32-chunk of `parse-info`s before doing `first` on them.

src/orchard/java.clj
```diff
@@ -476,8 +476,7 @@
   [ns sym]
   (when-let [ns (find-ns ns)]
     (->> (vals (ns-imports ns))
-         (map #(member-info (-> ^Class % .getName symbol) sym))
-         (filter identity)
+         (keep #(member-info (-> ^Class % .getName symbol) sym))
```
alexander-yakushev, line 479: This one is minor.

vemv:
> Still, maybe limiting it to a single Java version at least makes sense?

Not necessarily as Eastwood detects Java method deprecations - these vary per JDK

Plus, code evaluation can fail for arbitrary reasons  - the chosen JDK is an important factor, so more JDKs == more coverage

vemv:
Thanks much for the changes!

Hopefully this will make at least some difference.

I'll want to work on https://github.com/clojure-emacs/orchard/issues/211 sometime in the year.

PR #239, opened 2024-04-15
[clojuredocs] Protect cache filling with a lock

alexander-yakushev:
While profiling the test runs, I noticed that the function that parser Clojuredocs export spends significant time within `pmap` invocation. Looks like multiple threads are racing to compute the cache.

The change is unlikely to impact users anyhow, unless there is a situation with a slow machine like there was in Compliment and its own caching. In any case, this change wouldn't hurt.

### DIFF
--- a/src/orchard/clojuredocs.clj
+++ b/src/orchard/clojuredocs.clj
@@ -8,11 +8,17 @@
    [clojure.string :as string]
    [orchard.util.os :as os])
   (:import
-   (java.io IOException)
    (java.net URI)
-   (javax.net.ssl HttpsURLConnection)))
+   (javax.net.ssl HttpsURLConnection)
+   (java.util.concurrent.locks ReentrantLock)))
 
 (def cache (atom {}))
+(def ^:private ^ReentrantLock lock
+  "Lock to prevent concurrent loading and parsing of Clojuredocs data and writing
+  it into cache. This lock provides only efficiency benefits and is not
+  necessary for correct behavior as accessing atom that contains immutable data
+  structures is safe without a lock."
+  (ReentrantLock.))
 (def default-edn-file-url
   "https://github.com/clojure-emacs/clojuredocs-export-edn/raw/master/exports/export.compact.edn")
 (def cache-file-name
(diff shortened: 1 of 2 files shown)


src/orchard/clojuredocs.clj
```diff
@@ -53,17 +54,20 @@
         (finally
           (.disconnect conn))))))
 
-(defn load-docs-if-not-loaded!
-  "Load exported docs from bundled or cached file when no docs are loaded.
+(let [lock (ReentrantLock.)]
```
bbatsov, line 57: Isn't it a bit weird to have the lock around a definition? 

I'd also suggest adding your comment from the issue here for posterity's sake.

alexander-yakushev: + on the comment.

Does the "function-private global variable" pattern look weird to you? I thought it was pretty common.

vemv:
Seems a reasonable addition!

You might want to add similar locks throughout the ns

src/orchard/clojuredocs.clj
bbatsov: > Does the "function-private global variable" pattern look weird to you?

I haven't seen it much, but I get why you did it like this. It's fine by me.

alexander-yakushev:
> You might want to add similar locks throughout the ns

Right, I see now that there is another place where stuff gets loaded. Let me rethink and refactor this a bit.

alexander-yakushev:
Tests are passing now.

test/orchard/clojuredocs_test.clj
```diff
@@ -69,54 +68,53 @@
...
-      (with-redefs [docs/test-remote-url (constantly [false (IOException. "dummy")])]
-        (create-dummy-cache-file now)
-        (reset! docs/cache {})
-
-        (is (= now (.lastModified cache-file)))
-        (is (empty? @docs/cache))
-        (is (thrown? IOException (docs/update-cache! test-edn-file)))
-        (is (= now (.lastModified cache-file)))
-        (is (empty? @docs/cache))))))
+      (docs/clean-cache!)
```
vemv, line 110: Should this be protected by the lock?

alexander-yakushev: It may be, but at the same time, it doesn't have to. Setting the cache to an empty map is not really something we care about if it happens on multiple threads simultaneously.

vemv:
This seems reasonable to merge although my humble suggestion would be to make this as 'foolproof' as possible - these sections of code can get updated once every couple years so something simple would probably be appreciated - especially as we're dealing with locking.

alexander-yakushev:
I don't think that serializing all access to cache (including reads) is necessary. I'll try to expand the comment so that future editors understand better what is going on.

vemv:
I didn't suggest serializing reads - instead I wanted to prevent any maintainers from having to reason much - energies are a pretty scarce resource when there are 100 issues open at any given time.

A readwrite lock might help. And/or encapsulating things in some or other way.

But if you don't see a clear improvement that's fine too - no issue!

alexander-yakushev:
Given that I'm the first person to take a look at this namespace in 4 years, I can say yes, the maintainers' energies are safe here. 😆

vemv:
What I meant is, next time I or someone else has to touch this, maybe 4 years from now, will we scratch our heads?

PR #243, opened 2024-04-21
[inspector] Misc improvements

alexander-yakushev:
Nothing user-facing here except removing `eval-and-inspect`.

The controversial change of using vector for `:rendered` is in its own commit. No tests added because of Clojure's list/vector equality semantics.  Besides, `:rendered` still becomes a seq in the end for now, until this transformation is moved into cider-nrepl instead.


vemv (approved):
Thanks!

bbatsov:
Nice changes! Thanks!

PR #249, opened 2024-04-23
More post-1.9 cleanup

alexander-yakushev:
Some things that slipped past me in the 1.9 removal cleanup.

Also, remove handling of obscure `clojure.spec` namespace which only existed in some 1.9 alphas.

### DIFF
--- a/src/orchard/misc.clj
+++ b/src/orchard/misc.clj
@@ -6,8 +6,6 @@
    [clojure.string :as string]
    [orchard.util.io :as util.io]))
 
-(require 'clojure.core.protocols)
-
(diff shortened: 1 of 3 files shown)


bbatsov:
This has always been very funny to me:

```
;; - clojure.spec (released between Clojure 1.8 and 1.9, but never included in Clojure) ;;
;; - clojure.spec.alpha (renamed from clojure.spec and included in Clojure 1.9)         ;;
;; - clojure.alpha.spec (spec-2, the new experimental version)                 
```
Brilliant naming! 😆

PR #252, opened 2024-04-24
[inspector] Display count in the header for objects with known size

alexander-yakushev:
Similarly to how it was already done for arrays, Counted objects now also show their size upfront.


vemv (approved):
lgtm, thanks!

bbatsov:
The code looks good to me, although it seems there's a small spacing mistake that results in failing CI.

PR #254, opened 2024-04-29
Fix tests on Clojure 1.12.0-alpha10

alexander-yakushev:
New representation of primitive arrays in 1.12 is `byte/1` and so on.


vemv:
Thanks!

Issue #242, opened 2024-04-21
Inspector: remove redundant Java package prefixes


vemv:
When inspecting classes, often all fields/methods/... are prefixed with the fully-qualified class name.

Example:

```
Name: com.corp.Dog
Class: java.lang.Class

--- Interfaces:
  java.io.Serializable
  java.lang.Cloneable
  java.lang.Comparable
  org.apache.thrift.TBase

--- Constructors:
  public com.corp.Dog()
  public com.corp.Dog(com.corp.Dog)

--- Fields:
  public boolean com.corp.Dog.isTopStylist
  public static final java.util.Map com.corp.Dog.metaDataMap
  public java.lang.String com.corp.Dog.sub

--- Methods:
  public void com.corp.Dog.clear()
  public int com.corp.Dog.compareTo(java.lang.Object)
  public int com.corp.Dog.compareTo(com.corp.Dog)

```

This is particularly ugly in `com.corp.Dog.isTopStylist` - there isn't an actual thing named `com.corp.Dog.isTopStylist` - it's `isTopStylist` so removing the prefix would be both more readable and more correct.

There may be cases where there's a legit use case for rendering a prefix - I'm thinking interface implementation or method inheritance.

Even then, I believe the right thing would be to render the short name, and provide any extra info when going `down` that particular method.

vemv:
@alexander-yakushev you may want to give this one a shot while you're with other things Inspector

alexander-yakushev:
I see what's going on. First, the method gets printed honoring the `max-atom-length` (which is 150 by default), and truncated above that limit. **Then** it gets shortened by stripping the package prefixes, so you get a truncated string that is shorter than 150 characters. The behavior hasn't changed from before (apart from shortening).

alexander-yakushev:
I don't mind dancing around truncation limits a bit (serialize into a separate printer, shorten, then truncate). But I'm not sure this is needed – method signature that is 100+ characters long is not very useful.

vemv:
However it would seem simple enough to disable all trimming for Method/Field/... `Value`s?

PR #255, opened 2024-04-29
[inspector] Prettify printed members when inspecting a class object

alexander-yakushev:
Addresses #242.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#246](https://github.com/clojure-emacs/orchard/issues/246): Enable sibling navigation for arrays.
 * [#252](https://github.com/clojure-emacs/orchard/issues/252): Display count in the inspector header for objects with known size.
 * [#253](https://github.com/clojure-emacs/orchard/issues/253): Rework pagination and datafy.
+* [#242](https://github.com/clojure-emacs/orchard/issues/242): Remove redundant Java package prefixes in the inspector.
 
 ### Bugs Fixed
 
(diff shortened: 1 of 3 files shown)


alexander-yakushev:
It's called prettify but the approach is not pretty. Anyway, at some point if we want more customizations it will have to be rewritten. But this should work for now.

I didn't replace the `orchard.inspect` representation for methods, fields, and such because, on its own, it is useful for a single method to show which class it belongs to. It's just that the inspector provides plenty of context when it renders many methods of the same class, so it makes sense to only override it there.

vemv (approved):
Thanks much!

The approach seems perfectly fine for a long while.

vemv:
The prettification could also be applied when inspecting a Method

image

or Field:

image

(however if we do that then it's not clear that we're within the Long class. Maybe we could add an additional header?

alexander-yakushev:
Just prettyfying the `Value` field can be easily done ad-hoc. Adding a header would require overloading `inspect` for methods and fields and constructors, then reusing the `:default` implementation which handles them now. I'd rather do the former. We can also render `Value` for them in some third specific way, e.g. `java.lang.Long: public static int bitCount(long)`, but I'm not sure it would be any better than how it looks now.

vemv:
Doing the prettification and nothing else is OK for now - in the end, the user knows what he's inspecting, or can go back if he forgot

I'd personally like the header but no rush - just leaving a discussion here is fine

vemv:
`java.lang.reflect.Field<java.lang.Long>` or some other format can also be OK (assuming it doesn't affect navigation)

PR #256, opened 2024-04-30
[inspector] Allow customizing max-nested-depth

alexander-yakushev:
Option was ignored when there was no default value for it in the map.


vemv (approved):
✅

PR #257, opened 2024-04-30
Misc maintenance

alexander-yakushev:
Copying question from Slack:

_I just noticed that Orchard test matrix separately runs tests with and without the :provided profile, can you please confirm that this still has a purpose and which one?_

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,6 @@ version: 2.1
 defaults: &defaults
   working_directory: ~/repo
   environment:
-    LEIN_ROOT: "true" # we intend to run lein as root
     JVM_OPTS: -Xmx3200m -Dclojure.main.report=stderr
 
 # Runners for each major OpenJDK
@@ -104,23 +103,26 @@ jobs:
   util_job:
     description: |
       Running utility commands/checks (linter etc.)
-      Always uses Java11 and Clojure 1.11
+      Always uses latest Java LTS and Clojure release
     parameters:
       jdk_version:
-        description: Version of JDK to test against
         type: string
+        default: "openjdk21"
+      clojure_version:
+        type: string
+        default: "1.11"
       steps:
         type: steps
     executor: << parameters.jdk_version >>
     environment:
-      VERSION: "1.11"
+      VERSION: << parameters.clojure_version >>
     steps:
       - checkout
       - with_cache:
-          parser_target: "parser"
+          parser_target: "parser-next"
           cache_version: "util-job-v1"
           jdk_version: << parameters.jdk_version >>
-          clojure_version: "1.11"
+          clojure_version: << parameters.clojure_version >>
           steps: << parameters.steps >>
 
   deploy:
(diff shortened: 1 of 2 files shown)


vemv:
IIRC the purpose is ensuring that Orchard is able to run in both the presence and absence of the `clojurescript` dependency.

Some code conditinally obseves the presence of cljs, so by maintaining this matrix, we hit both branches of the code.

alexander-yakushev:
Given that CLJS features that Orchard provides don't intersect with anything that Java brings and with javadoc parsing, would it make sense to trim the matrix there? Pick a single parser type for cljs tests and the latest JDK (or oldest, or both). This can save us ~20 testing jobs per run.

vemv:
I'd be OK with shaving parsing stuff, but I'd still run all JDK<->cljs combinations, out of caution.

alexander-yakushev:
How about doing it the other way around? Test everything with CLJS included, but have a couple sanity-check jobs without it on the classpath. Not having it on classpath only tests that Orchard requires it conditionally, right? Should be enough to do it once or twice.

vemv:
Depends on the nature of the sanity check. I'd be open to check it out but, please be advised, I tend to lean towards safety.

In practice, the CI matrix hasn't been a bottleneck since its introduction some 3y ago.

Normally if one job fails/passes, that's all the info I need and I can focus elsewhere till it's all complete (which is parallel).

(plus, one has to wait for a code review anyway)

In return we get a fair bit of peace of mind - sometimes we get pretty odd issue reports, so ruling out some basics helps in quickly diagnosing user reports as they happen.

alexander-yakushev:
Circle begins to throttle parallelism when pushing code actively, so it gets mildly annoying. Safety is nice, but it also doesn't make sense to waste cycles just for the sake of it. GIven the multiplicative nature of CI matrix dimensions, there has to be a limit where the spread has to stop.

Do you have off top of your mind an example of a report that was caused by a weird combination of parameters? That would certainly change my mind.

alexander-yakushev:
Last commit is an example how it can look like. Brings it to 55->34 jobs.

alexander-yakushev:
Another point that I have previously brought up is Eastwood on every matrix cell. Last time you've mentioned that because it loads code, then running it CLJxJDK makes sense, and I agree. But doing that for every parser combination is wasteful (since it lints namespaces all the same).

vemv:
>  But doing that for every parser combination is wasteful (since it lints namespaces all the same).

👍 good to slash that part

> Circle begins to throttle parallelism when pushing code actively, so it gets mildly annoying. Safety is nice, but it also doesn't make sense to waste cycles just for the sake of it. GIven the multiplicative nature of CI matrix dimensions, there has to be a limit where the spread has to stop.

I'd just really appreciate exercising the JDK*cljs part. It helps me as maintainer to rule out that possible source of issues.

The matrix has helped us out in many forms. Asking for a specific example is too much of an ask (when any given day there many better uses of my time), suffice to say that very few people to actively maintain the cljs part, so let's try to find middle grounds.

The build is pretty quick and if you commit N times in a row, it's easy enough to go over CCI and stop older commits.

In general, let's try to balance out dev convenience vs. user-facing correctness. Neither extreme is the best choice.

Cheers - V

vemv (approved):
Great work!

I appreciate the fine-grained trimming.

For the record, it's indeed a good concern to optimize these matrices. CI providers make it too easy for things to explode combinatorially - ideally they'd offer a better syntax.

PR #260, opened 2024-05-01
[print] Prevent writer going over the total limit

alexander-yakushev:
The bug is minor (in certain scenarios the printer can overflow the limit by <5%), so `0.24.0` is still good enough for cider-nrepl.


alexander-yakushev:
Also reduce :max-value-length for inspector to 10000, because 50000 is still too much for Emacs too handle on the single line (it works but is slow as hell).

vemv:
Thanks!

PR #265, opened 2024-05-03
[inspector] Make spacious printing customizable

alexander-yakushev:
Still keeping spacious=true by default in Orchard to avoid rewriting the tests for now, in the unlikely case we'll have to revert it back.


vemv (approved):

alexander-yakushev:
When you have time, please deploy `0.24.1`, so I can make a cider-nrepl PR, thanks!

PR #267, opened 2024-05-03
[inspect] Don't truncate member names

alexander-yakushev:


vemv:
Awesome - thanks much!

PR #270, opened 2024-05-14
[inspector] Use preconditions that throw exceptions instead of errors

alexander-yakushev:
Misconfiguration throwing AssertionErrors makes them slip through cider-nrepl's `with-safe-transport`.

Overall, assertions are bad when integrating with other code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## master (unreleased)
 
 * [#267](https://github.com/clojure-emacs/orchard/issues/267): Inspector: don't truncate constructor/field/method names.
+* [#269](https://github.com/clojure-emacs/orchard/issues/269): Inspector: add `refresh` function as a single entrypoint to changing inspector config.
+* [#270](https://github.com/clojure-emacs/orchard/issues/270): Inspector: throw exceptions instead of AssertionErrors when inspector is misconfigured.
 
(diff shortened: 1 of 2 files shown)


vemv:
I'd be happy as well with fixing with-safe-transport - seems fairly hard to guarantee across projects/time that we'll never use assertions.

alexander-yakushev:
Agreed – additionally handling assertion errors is reasonable in `with-safe-transport`, but not blanketly catching Throwable (which is a bad idea).

bbatsov:
Do we even need those preconditions?

alexander-yakushev:
I'd say yes. For example, setting a zero for certain values can hang the inspector of raise very low-level errors. Since those values can ultimately come from the user, we have to validate them somewhere. And I don't like the idea of validating them on each layer (orchard, cider-nrepl, cider); I'd rather do that in a single place that knows what the values should look like (orchard) and flow the helpful errors back through the other layers.

vemv (approved):
The assertion system isn't best for validating user input (since users can disable it, for one thing) so this seems a sensible change.

The cider-nrepl improvement would seem welcome as well.

bbatsov:
I just assumed that bad data would result in exceptions, not lock-ups. If that's not the case - I guess we need those assertions. In general I've rarely found them useful in my work with Clojure, that's why I'm somewhat skeptical of their value.

PR #271, opened 2024-05-14
[inspector] Introduce object view mode

alexander-yakushev:
This is a feature that I've been wanting for a long long time. The idea is to have a toggle button (e.g. `v`) that forces inspector to render the current value as an unrecognized Java object – i.e., by displaying its fields.

image <br>

image  <br>

image  <be>

This is very useful when working on Clojure itself or developing custom data structures that happen to implement List, Map, etc., but you want to see the implementation bits.

There are a couple of details I have not decided on yet:

1. Originally, I wanted this to be a toggle (normal mode <-> java mode). However, this has a potential to add more view modes. I'm not sure which ones those would be right away, but I can possibly imagine something like simple mode (don't show metadata and datafy – for cases where those are too large and distract from viewing the value), a "pretty mode" that does uses pretty printing where currently there is a regular printer, etc.
2. If there are more modes, then the "toggle" button can cycle through them. However, I don't think such cycling should be implemented on Orchard side, I think having a `:view-mode` as a plain config value is sufficient at this level. CIDER or cider-nrepl can then tackle user interactions (cycling, dedicated switches, both).
3. I'm not set on the names of view modes (currently `:normal` and `:java`). Bikeshedding is welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [#267](https://github.com/clojure-emacs/orchard/issues/267): Inspector: don't truncate constructor/field/method names.
 * [#269](https://github.com/clojure-emacs/orchard/issues/269): Inspector: add `refresh` function as a single entrypoint to changing inspector config.
 * [#270](https://github.com/clojure-emacs/orchard/issues/270): Inspector: throw exceptions instead of AssertionErrors when inspector is misconfigured.
+* [#271](https://github.com/clojure-emacs/orchard/issues/271): Inspector: introduce object view mode.
 
 ## 0.25.0 (2024-05-03)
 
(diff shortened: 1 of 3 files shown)


vemv (approved):
Great addition - appreciated!

alexander-yakushev:
Renamed `:java` to `:object` – that probably makes more sense.

bbatsov:
> If there are more modes, then the "toggle" button can cycle through them. However, I don't think such cycling should be implemented on Orchard side, I think having a :view-mode as a plain config value is sufficient at this level. CIDER or cider-nrepl can then tackle user interactions (cycling, dedicated switches, both).

Agreed, that should be on the clients to implement.

> I'm not set on the names of view modes (currently :normal and :java). Bikeshedding is welcome.

The `:object` name you chose makes sense to me. Something like `raw` could also work. I'm afraid I'm not particularly inspired with the names today.

PR #274, opened 2024-05-31
[print] Remove support for spacious printing

alexander-yakushev:
Leaving this unmerged for now, let it just hang until we confirm that people are OK with the new non-spacious printing being the default.


alexander-yakushev:
As expected, nobody is noticing this change. Let's remove it.

PR #275, opened 2024-06-01
[inspect] Re-implement :view-mode with a stack same as paging

alexander-yakushev:
In the end, I decided to track `:view-mode` on the Orchard side and make it behave as a per-value remembered thing similar to how current-page is implemented (like I said in https://github.com/clojure-emacs/cider-nrepl/pull/880).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## master (unreleased)
 
-## 0.26.0 (2024-06-30)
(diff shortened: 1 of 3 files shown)


vemv (approved):

vemv:
Makes sense to me!

alexander-yakushev:
`0.26.1` would be much appreciated to finish the cider-nrepl side.

PR #280, opened 2024-07-19
[print] Fix bug when a printed collection is not Iterable

alexander-yakushev:
This is a fix for the issue raised on Slack: https://clojurians.slack.com/archives/C0617A8PQ/p1721356711913299

I've been blindly calling `.iterator` on any Clojure collection in the printer, but not everything that implements IPersistentVector or IPersistentSet is `Iterable`.

In another commit, I rewrite `TruncatingStringWriter`, which used by the printer, to throw an error when the limit is riched, in the same manner as nREPL's `QuotaBoundPrinter` works. This simplifies the implementation of `print-coll`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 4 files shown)


bbatsov (approved):
Looks good to me.

PR #282, opened 2024-08-01
[inspect] Don't crash when inspecting internal classes

alexander-yakushev:

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -575,11 +575,14 @@
 (defn- shorten-member-string [member-string, ^Class class]
   ;; Ugly as hell, but easier than reimplementing all custom printing that
   ;; java.lang.reflect does.
-  (-> member-string
-      (string/replace #"[\w\.]+\.(\w+\()" "$1") ;; remove class from method name
-      (string/replace #"[\w\.]+\.(\w+)$" "$1") ;; remove class from field name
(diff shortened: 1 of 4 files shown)


src/orchard/inspect.clj
```diff
@@ -575,11 +575,13 @@
...
   ;; java.lang.reflect does.
-  (-> member-string
-      (string/replace #"[\w\.]+\.(\w+\()" "$1") ;; remove class from method name
-      (string/replace #"[\w\.]+\.(\w+)$" "$1") ;; remove class from field name
-      (string/replace (.getCanonicalName class) (.getSimpleName class))
-      (string/replace #"java.lang.([A-Z])" "$1")))
+  (as-> member-string s
+    (string/replace s #"[\w\.]+\.(\w+\()" "$1") ;; remove class from method name
+    (string/replace s #"[\w\.]+\.(\w+)$" "$1") ;; remove class from field name
+    (if-let [canonical (.getCanonicalName class)]
```
bbatsov, line 581: Maybe add a comment here explaining when we won't have the canonical name?

bbatsov:
The PR looks good to me.

src/orchard/inspect.clj
alexander-yakushev: Left a comment, the list is in the javadoc for `.getCanonicalName`.

PR #283, opened 2024-08-01
[spec] Cleanup 1.8 compatibility code

alexander-yakushev:


bbatsov:
I really hope that spec-1 will die at some points, because the current situation is just sad/weird. Thanks for tackling this!

PR #285, opened 2024-08-15
Remove special handling of Boot

alexander-yakushev:
This PR removes handling of Boot's classpath magic. This would not make CIDER "broken" with Boot, the REPL will still start and work (if somebody manages to run Boot, that is), but the "jump to source" functionality will become wobbly.

---


bbatsov:
No objections from me.

PR #286, opened 2024-08-15
Simplify directory setup and CI

alexander-yakushev:
1. Parser separation in different directories doesn't seem to be necessary. The files are required conditionally depending on the JDK version, and they are all packaged together in the end anyway.
2. Test namespaces for JDK9+ can also live with other test files in `test/` with some tweaks to make those tests not run on JDK8.
3. Some linter restructuring, nothing major.
4. **Reduce number of assertions and fix java test**. `java-test` had a bug that instead of testing each class imported into the namespace it tested `clojure.lang.Compiler` a hundred times. I also refactored the test to do bigger-grain assertions so that the output in case of failure is not completely overwhelming.

## Untie enrich-classpath from testing and download JDK sources

This is the biggest change. Turns out, Adoptium hasn't been including `src.zip` in the Docker images for quite a while. So all that extensive and all-covering testing with different parsers actually didn't test anything because there were no sources on the classpath. This PR:
- Downloads JDK sources in the Makefile as one of the preparation steps and repackages them.
- Puts downloaded sources onto the classpath in `project.clj`.
- Removes enrich-classpath from the testing pipeline as it's no longer necessary to test Orchard.

Since I've only made a script for downloading JDK21 sources, that's why only that JDK version is tested with different parsers. Packaging other JDK sources to test other JDKs can be done separately if somebody wants to do that.

---

- [x] All tests are passing


alexander-yakushev:
The idea behind the updated testing matrix:

1. We have JDK sources for JDK21. So, for that JDK version, test Orchard with sources for all Clojure versions and both for regular parser and parser-next.
2. For all other JDK versions, test for all sources and for the default parser (which would be legacy-parser in JDK8 and regular parser in other JDKs. The selection happens automatically inside Orchard.)
3. Have a sanity test without Clojurescript and spec-alpha2 on the classpath. THis checks that there are no hard requirements on those classes existing. One version tuple is enough for this.
4. Eastwood is run for the latest stable Clojure version and for the oldest and newest JDK.

bbatsov:
I'm fine with the proposed changes. It will probably be a good idea to document the CI matrix, so it's easier for people to figure out why we've structured it as we did. (e.g. something similar to your last comment on this PR)

alexander-yakushev:
Added some comments to the CI yaml file.

bbatsov:
Thanks!

PR #287, opened 2024-08-16
[inspect] Don't crash when field contains non-equiv()able value

alexander-yakushev:
Happens when the object declares but doesn't implement all the necessary methods. Such object is not correct, but we should still be aware.


bbatsov:
The change looks good. I think it deserves a mention in the changelog, as it's a bugfix more or less.

PR #288, opened 2024-08-20
[java] Don't crash if none of the parsers can be initialized

alexander-yakushev:
Context here: https://clojurians.slack.com/archives/C17JYSA3H/p1724172844138059

On the JRE, loading CIDER (and hence Orchard) produces an exception like this:

```
error in process sentinel: Could not start nREPL server: Downloading: cider/cider-nrepl/0.45.0/cider-nrepl-0.45.0.pom from clojars
Downloading: cider/orchard/0.22.0/orchard-0.22.0.pom from clojars
Downloading: mx/cider/logjam/0.2.0/logjam-0.2.0.pom from clojars
Downloading: cider/orchard/0.22.0/orchard-0.22.0.jar from clojars
Downloading: mx/cider/logjam/0.2.0/logjam-0.2.0.jar from clojars
Downloading: cider/cider-nrepl/0.45.0/cider-nrepl-0.45.0.jar from clojars
Execution error (IllegalArgumentException) at orchard.misc/require-and-resolve (misc.clj:161).
No such namespace: orchard.java.parser-utils
```

---

- [ ] You've updated the [changelog](../blob/master/CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## master (unreleased)
 
 * [#285](https://github.com/clojure-emacs/orchard/issues/285): **BREAKING:** Remove special handling of Boot classpath.
-* [#287](https://github.com/clojure-emacs/orchard/issues/287): Inspector: ton't crash when field contains non-equiv()able value.
+* [#287](https://github.com/clojure-emacs/orchard/issues/287): Inspector: don't crash when field contains non-equiv()able value.
+* [#288](https://github.com/clojure-emacs/orchard/issues/288): Java: don't crash when loading if none of the parsers could be correctly initialized.
 
 ## 0.26.3 (2024-08-14)
 
--- a/project.clj
+++ b/project.clj
@@ -110,7 +110,8 @@
                          :eastwood {:ignored-faults {:unused-ret-vals-in-try {orchard.java {:line 84}
                                                                               orchard.java.parser-next-test true}}
                                     :exclude-namespaces ~(if jdk8?
-                                                           '[orchard.java.parser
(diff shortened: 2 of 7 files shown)


alexander-yakushev:
The culprit was not the `source-info` triggering parser loads, but the `def module-info` line requiring `parser-utils`. Given that `module-info` was loaded a bit awkwardly already (and there was a comment affirming it), I've moved it into a separate namespace that can be safely loaded on any JDK after 8.

Another sidenote is the usage of JDK9 in the names. Obviously, we won't change the existing ones for no reason, but I suggest we use JDK11 as the LTS version that we actually support.

bbatsov:
I'm fine with updating the references to JDK 9 - they are there just because it introduced the module system (that probably no one is using still) and all the complexity that came with it. I think the LTS concept was added shortly afterward, but we forgot to update the references accordingly.

src/orchard/java/jdk11.clj
```diff
@@ -0,0 +1,7 @@
+(ns orchard.java.jdk11
```
bbatsov, line 1: I doubt there will be anything else here that's not related to modules, so this can also be something like `orchard.java.modules`. I'm not very fond of linking ns names to specific versions of something. But I don't feel strongly about this.

alexander-yakushev:
Replaced all mentions of JDK9.

bbatsov:
This might be a good time for a new Orchard release.

theronic:
Which version of cider-nrepl includes this fix, please?

alexander-yakushev:
The latest one certainly does.

PR #292, opened 2024-08-28
[print] Fix map entries being spliced in non-map collections

### LINKED ISSUE #291: orchard.print renders map entries in a misleading way
## Expected behavior
Map entries should be printed as 2-vectors. 

## Actual behavior

```
(orchard.print/print-str (seq {:a 1 :b 2})) ;; => "(:a 1 :b 2)"
;; expected: ([:a 1] [:b 2])
```
This affects Cider's debugger and inspector, and can be traced to the `:map-entry` dispatch of orchard.print/print:
https://github.com/clojure-emacs/orchard/blob/ed119f0075eacb951012763fccd8afa1a9fb1975/src/orchard/print.clj#L95

alexander-yakushev:
Fixes #291.

---


src/orchard/print.clj
```diff
@@ -43,8 +42,18 @@
...
+  [^TruncatingStringWriter w, x, map?]
+  (if (and map? (instance? Map$Entry x))
+    (do (print (.getKey ^Map$Entry x) w)
+        (.write w " ")
+        (print (.getValue ^Map$Entry x) w))
+    (print x w)))
+
 (defn- print-coll [^TruncatingStringWriter w, ^Iterable x, ^String sep
-                   ^String prefix, ^String suffix]
+                   ^String prefix, ^String suffix, map?]
```
bbatsov, line 56: Not a big fan of this boolean param, but I can live with it. Perhaps we can at least make it somewhat optional, so we don't have to pass `false` around.

alexander-yakushev: Updated it to be optional. Does it look that much better for you? It's a private function anyway.

bbatsov: Not much better, but a bit better - in general I'm not a big fan of boolean params, as I always have to consult the method signature to remember/understand what some boolean flag is supposed to do. But I agree it's not a big deal in our case, given it's a private API.

bbatsov (approved):

yuhan0:
Thanks for the quick fix! Note there's a typo in the changelog, 290 -> 292

Issue #278, opened 2024-06-17
cider-doc for java methods with array args rendered incorrectly


bsless:
## Expected behavior

Call `cider-doc` on `java.nio.file.Path/of`, render docs such that `more` is represented as an array:
```
java.nio.file.Path/of
^Path [^String first, ^String<> more] ; or  [Ljava.lang.String;
^Path [^java.net.URI uri]
```

## Actual behavior

```
java.nio.file.Path/of
^Path [^String first, ^String more]
^Path [^java.net.URI uri]
```

## Steps to reproduce the problem

`M-x cider-doc` `java.nio.file.Path/of`

## Environment & Version information

```
Cider version 1.15.0
emacs --version
GNU Emacs 30.0.50
Development version 319e16550d66 on master branch; build date 2024-05-31.
```

### Clojure version

E.g. 1.11.1

### Java version

21

### Operating system

Pop OS 22.04

vemv:
Thanks!

Some notes https://clojurians.slack.com/archives/C0617A8PQ/p1718605593241419

alexander-yakushev:
As far as I can tell, this is already fixed in orchard 0.28.0/cider-nrepl 0.50.3 with #295.

<img width="368" alt="image" src="https://github.com/user-attachments/assets/9b5e96e7-9228-4e91-9e4d-18c2bd7fed6d" />

Thank you for the report, @bsless!

PR #295, opened 2024-10-23
[java] Preserve array indicator in member docs and remove reliance on --add-opens in parser-next

alexander-yakushev:
Addresses #278.

The issue is quite confusing. Clojure's `clojure.reflect` namespace when dealing with square brackets `[]` transforms them to `<>` for whatever reason. In some places, it then transforms it back, e.g. in function `clojure.reflect/typesym`.

Orchard in different places removes `<>` indicators, possibly presuming they refer to generic arguments. However, with reflection, you can't get generics information anyway, since generics are not a runtime feature.

This PR ensures that things like `java.lang.String<>` are transformed back to `java.lang.String[]`, and also doesn't do any "generics removal" for reflection-obtained docs.

UPDATE: This PR now also includes the rewrite of parser-next to use public interfaces instead of internal closed-module implementations. It's hard to merge these two big changes separately because some of the tests are inter-reliant.


bbatsov (approved):
The changes look good to me. Great work!

alexander-yakushev:
@vemv Just FYI.

PR #297, opened 2024-10-24
Test parsers on all JDKs >=11

alexander-yakushev:
Before merging the parser implementations, it is a good idea to restore the parser testing for all JDK versions (except for JDK8 for now). I've updated the relevant scripts to download and repackage JDK sources for all JDK targets we run tests for.


bbatsov (approved):
Looking good!

PR #298, opened 2024-10-25
[java] Remove orchard.java.parser

alexander-yakushev:
Since parser-next should now work in all conditions that parser used to work, the latter is no longer needed. This PR removes it completely. There were no interesting testcases for it that are not already covered by parser-next, so I removed its testcase namespace too.

@vemv fyi, just so you know what's going on.


bbatsov (approved):
Fantastic work, Sashko! 👏

PR #299, opened 2024-10-30
[ci] Simplify deploy script

alexander-yakushev:
The current pipeline was quite convoluted by going CircleCI->custom Clojure script->Makefile->Leinigen. I've cut the custom script from the chain and replaced it with the preparatory actions inside Makefile.


Makefile
```diff
@@ -73,6 +73,11 @@ lint: kondo cljfmt eastwood
 # Deployment is performed via CI by creating a git tag prefixed with "v".
 # Please do not deploy locally as it skips various measures.
 deploy: check-env clean
+	@if ! echo "$(CIRCLE_TAG)" | grep -q "^v"; then \
+		echo "CIRCLE_TAG: $(CIRCLE_TAG) must start with 'v'."; \
```
bbatsov, line 77: Probably we can prefix this with `[Error]`, so it's easier to spot it in the CI logs.

bbatsov:
I agree that's simpler. We should do the same for `cider-nrepl` as well. I think those are the only repos that are currently doing auto-deployments.

PR #301, opened 2024-12-03
[java] Remove orchard.java.legacy-parser

alexander-yakushev:
This drops support for Java sources parsing on JDK8.


bbatsov:
Looks good overall! Maybe we should also add some `Caveats` or `Java 8 support` section in the README where we mention this, so it's obvious we've soft deprecated Java 8 and some features are not available for it.

PR #302, opened 2024-12-03
[java] Merge orchard.java.parser-utils and parser-next

alexander-yakushev:
`parser-utils` namespace was extracted because it was reused between `parser` and `parser-next`. The separation is not needed anymore, so I'm merging them back. It kinda sucks that we lose good Git history with this, but it's already confusing as it is with three namespaces where stuff happens (`orchard.info`, `orchard.java`, `orchard.java.parser-next`).


bbatsov (approved):

bbatsov:
Agreed. I think the trade-off with the git history is fine.

bbatsov:
I am also thinking that `parser-next` can become simply `parser` down the road.

Issue #261, opened 2024-05-01
Support for spec-alpha2?


alexander-yakushev:
It looks like the development of clojure spec-alpha2 (also widely and hilariously known as `clojure.alpha.spec`) is stalled for now (as confirmed by Alex Miller too: https://clojurians.slack.com/archives/C03S1KBA2/p1679924331305749?thread_ts=1679878261.044279&cid=C03S1KBA2). The library doesn't have a JAR release and not even a single git tag.

Orchard supports it nominally in `orchard.spec`.

Two questions:
1. How important is having support for an almost-abandoned alpha library?
2. If the support for it desired, how important is to test against it? We currently have to clone that repository in each CI job for dubious benefits.

vemv:
Apparently the Spec browser supports Spec 2 https://docs.cider.mx/cider/usage/misc_features.html#browsing-the-clojure-spec-registry

I also know of misc professional projects that use Spec 2.

Perhaps a matrix trimming might be welcome (as long as that work isn't burdensome to you), otherwise I haven't observed a practical problem.

Worth noting, we always encourage running `make test` as the standard command across projects, so we shouldn't be too concerned about what's inside.

alexander-yakushev:
> I also know of misc professional projects that use Spec 2.

😱 

It's just a bit annoying having to do a special dance for a single library that never left alpha to ensure that the only one unit test written for it works.

vemv:
Some low-hanging fruit is that if locally one runs `make test` twice, one is `clean`ing twice so the clone will be performed twice.

vemv:
We also could use lein-git-down, it's a "xs" task, can do for you if you consider things painful locally

(Personally I don't)

bbatsov:
@alexander-yakushev Looking at the "State of Clojure 2024" results I don't see much about spec. There's one question showing it's the least important aspect of Clojure for most people https://www.surveymonkey.com/results/SM-hht04mGydwZ6Nqr7N8vjCA_3D_3D/ I thinking that we can definitely trim some spec-related functionality, but let's see what our Survey will unveil as well. 

<img width="537" alt="image" src="https://github.com/user-attachments/assets/7ef326c2-4f71-470f-adc4-98ba0b4032aa" />

PR #308, opened 2025-01-03
Testing and project root cleanup

alexander-yakushev:
Mostly inconsequential changes aimed to reduce noise in the root dir and remove unimportant parts.

- I removed `dev/user.clj` because its contents look to me like something that should be set up system-wide, not specific to the project. The developer is free to include tools.namespace in the `:user` profile and use it. Besides, CIDER itself ships with both tools.namespace and clj-reload.
- I removed the spec2 submodule for the reasons first mentioned in #261. The support for Spec2 remains intact; and we never touch that code to break anything. It's just that testing that part involves an extra song and dance that's just not worth it much. If Spec2 ever gets an actual release, we can review this decision.

I'm OK with this hanging for a while, until we get the final results on the State of CIDER survey.


bbatsov (approved):
I like all the proposed changes. I have to admit I have forgotten about all those things, otherwise I would have changed them earlier.

PR #310, opened 2025-01-05
[java] Download 3rd-party Java sources from Maven

alexander-yakushev:
Finally, a way to download source JARs at runtime. Sister PRs: https://github.com/clojure-emacs/cider-nrepl/pull/911, https://github.com/clojure-emacs/cider/pull/3769.

I tried a few approaches and settled with this one – downloading happens within `class->source-file-url` if `*download-sources-jar-fn*` is bound. The caller is expected to bind it to something that eventually invokes `download-sources-jar-for-coordinates` but can do custom logic and predicates too. I'm not 100% happy with this solution, it feels kinda awkward, but this is the best I could do. I honestly tried to make it strictly "cider-nrepl invokes Orchard", but the amount of repeated work was too much. Common Lisp condition system would actually make this cleaner, but hey, that's what dynamic vars actually are in the end, right?

Different ways of downloading is supported. Ultimately, all of them involve spawning subprocesses:

- Clojure 1.12 projects with `clojure` build tool can invoke the new shiny `invoke-tool` function.
- Pre-1.12 you can subshell `clojure` directly.
- Leiningen projects can call `lein`.

Orchard doesn't do build tool detection on its own, the caller is expected to pass it explicitly.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
+* [#310](https://github.com/clojure-emacs/orchard/pull/310): Java: add functions for downloading 3rd-party Java sources from Maven.
 * [#309](https://github.com/clojure-emacs/orchard/pull/309): **BREAKING:** Remove deprecated functions from orchard.java (`jdk-find`, `jdk-sources`, `jdk-tools`, `ensure-jdk-sources`).
 * [#309](https://github.com/clojure-emacs/orchard/pull/309): **BREAKING:** Remove `orchard.java/cache-initializer`.
 * [#311](https://github.com/clojure-emacs/orchard/pull/311): Trace: fix the printing inside the wrapped function to be truncated.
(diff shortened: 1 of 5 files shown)


bbatsov:
> I'm not 100% happy with this solution, it feels kinda awkward, but this is the best I could do. I honestly tried to make it strictly "cider-nrepl invokes Orchard", but the amount of repeated work was too much. Common Lisp condition system would actually make this cleaner, but hey, that's what dynamic vars actually are in the end, right?

Yeah, it's not work of art, but I think the current implementation is fine, and we can always polish it down the road as I doubt it will have other users besides cider-nrepl for the foreseeable future. 

Perhaps it'd be good to add some section to the README explaining how clients are supposed to use `*download-sources-jar-fn*`.

alexander-yakushev:
Expanded the README.

Also, moved the discovery of the underlying build tool to here. It is more reliable to do on the Clojure side.

I don't rush to merge this; I will start merging once I get approval on all three PRs.

bbatsov (approved):

PR #313, opened 2025-01-10
[java] Return actual paths as :file in class-info*

alexander-yakushev:
Context here: https://clojurians.slack.com/archives/CBE668G4R/p1735807627304719

Cider-nrepl's `info` doc says that `:file` key contains either an absolute URL or a relative path to the file – but still a filesystem path. The current class-info* implementation returns `:file` as a path on the classpath. Sometimes, it is a valid as a file path – if the class is local to the project, but more often the class is external, and the resource path doesn't work as the file path.

This PR fixes that.

---

### DIFF
--- a/src/orchard/java.clj
+++ b/src/orchard/java.clj
@@ -277,8 +277,13 @@
           source-file-url (src-files/class->source-file-url c)
           result (misc/deep-merge (reflect-info (reflection-for c))
                                   (when source-file-url
-                                    {:file relative-source-path
-                                     :file-url source-file-url})
+                                    ;; :file is the original historic key.
+                                    ;; :file-url was added later and is used by
+                                    ;; CIDER more now. Both have the same value
+                                    ;; for compatibility.
+                                    {:file     source-file-url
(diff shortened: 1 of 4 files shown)


bbatsov:
Yeah, that's fair. We've noticed a while ago that `:file` was somewhat problematic in its current form, but at the time we've opted to introduce `:file-url` instead of changing it. But I don't think anything will break by this change and it definitely makes more sense as the return value.

src/orchard/java.clj
```diff
@@ -277,8 +277,9 @@
           source-file-url (src-files/class->source-file-url c)
           result (misc/deep-merge (reflect-info (reflection-for c))
                                   (when source-file-url
-                                    {:file relative-source-path
-                                     :file-url source-file-url})
+                                    {:file     source-file-url
```
bbatsov, line 280: You might want to add a note here that both keys are essentially the same for historical reason.s

alexander-yakushev: Done

bbatsov (approved):

PR #315, opened 2025-02-24
[ci] Enable Windows tests

alexander-yakushev:
- [ ] You've updated the [changelog](../blob/master/CHANGELOG.md)


bbatsov (approved):
Good work. I'm not sure how many Windows users we have, as we've never tested much on it (I assumed most Emacs users there are using WSL), but I think that's never a bad idea.

alexander-yakushev:
Yeah, I don't think it will be much work to continue supporting Windows. We can always opt out of supporting particular advanced features, but it is good to know that we at least don't accidentally break some core workflows.

PR #317, opened 2025-03-04
Remove deprecated functions and refactor some tests

alexander-yakushev:
This PR removes most of the older deprecated functions. cider-nrepl no longer uses them, and I doubt anybody else does.


bbatsov:
Good changes!

PR #318, opened 2025-03-04
Remove unused orchard.misc functions

alexander-yakushev:
Further cleanup of stuff we no longer need. Most of these were necessary when we supported pre-1.10, but they are not used anymore.

I don't fret making breaking changes here since I don't think that relying on a misc namespace of a 3rd-party library is such a good idea anyway.


bbatsov (approved):

bbatsov:
Yeah, I'm not particularly concerned about any of those myself. And it'd be good to clean all of the cruft before getting to 1.0.

PR #320, opened 2025-03-05
More permissive orchard.info parsing of stacktrace output

alexander-yakushev:
This is part 1 of simplifying Haystack usecases. I think this PR is self-sufficient enough to be reviewed individually.

The proposed changes relax the requirements for Java classes/members and Clojure vars to be inferred using `orchard.info/info`. Now it can recognize all printed stacktrace formats. The examples are in the tests. This influences:
- Better jump to definition (can jump directly from the REPL output without no extra prep)
- Better cider-doc support
- Even eldoc (this looks a bit funny, consider yourself if we should restrict this somehow).

image

image

image

image

image

image

---


bbatsov (approved):
The proposed changes look very good to me!

I'd say the eldoc in the stacktraces is a feature.

PR #321, opened 2025-03-06
Revert clojure.string alias back to str

alexander-yakushev:
I know this may look petty, but I lost count how many times I've typed `(str/`, pressed complete, figured that the namespace probably doesn't have clojure.string required yet, going to the ns form only to discover it's `string` in this project. I'm not a fan of useless churn, but this is seriously annoying. Let's agree that `str/` is an "industry standard" and forget about this.


bbatsov:
No objections from me. I think we had changed it for consistency, but I'm also used to using `str/`, so I won't miss the `string` alias.

alexander-yakushev:
It happened in #171 and I wasn't too happy about that then either 😅.

PR #322, opened 2025-03-07
[stacktrace] Repatriate stacktrace analyzer from Haystack

alexander-yakushev:
This PR brings `haystack.analyzer` back home to Orchard. This is not an exact copy; I also did a lot of work in rewriting it to make leaner and faster.

- Pre-1.10 support bits are removed.
- Complete rework of :project flag for frames. Previously, it was implemented by scanning the classpath for all loaded namespaces, and comparing frame's namespace against that set. Now, frame namespace tries to be resolved to classpath resource. If successful, and the path starts with project's workdir, the frame is considered :project. Ridiculously faster this way (< 1 ms vs ~100 ms in a big project).
- Many functions were refactored without changing the behavior. 
- Test namespace was ported almost without behavior changes (but was rewritten to matcher-combinators).

The next step will be to switch cider-nrepl to this. I suspect that as I start working on that, I'll have to make some fixes here too; but I think it's fine to review and merge the current state first.

---

### DIFF
--- a/src/orchard/stacktrace.clj
+++ b/src/orchard/stacktrace.clj
@@ -1,0 +1,317 @@
+(ns orchard.stacktrace
+  "Instruments for analyzing exceptions and stacktraces which process exception
+  objects and attach extra data to them."
+  {:added "0.31"
+   :author "Jeff Valk, Oleksandr Yakushev"}
+  (:require
+   [clojure.java.io :as io]
+   [clojure.pprint :as pp]
+   [clojure.repl :as repl]
+   [clojure.spec.alpha :as s]
+   [clojure.string :as str]
+   [orchard.info :as info]
+   [orchard.java.resource :as resource])
+  (:import
+   (java.io StringWriter)
+   (java.net URL)
+   (java.nio.file Path)))
+
+(def ^:private ^Path cwd-path (.toAbsolutePath (.toPath (io/file ""))))
+
+(defn- pprint-write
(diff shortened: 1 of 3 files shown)


src/orchard/stacktrace.clj
```diff
@@ -0,0 +1,318 @@
+(ns orchard.stacktrace
+  "Cause and stacktrace analysis for exceptions"
+  {:added "0.31"
```
bbatsov, line 3: Maybe you can expand this a bit, to be more in line with the level of high-level descriptions that most other namespaces have.

test/orchard/stacktrace_test.clj
```diff
@@ -0,0 +1,342 @@
...
+  (testing "Exception cause unrolling"
+    (is (= 2 (count causes1)))
+    (is (= 1 (count causes2))))
+  (testing "Exception data"
+    ;; If ex-data is present, the cause should have a :data attribute.
+    (is (:data (first causes1)))
+    (is (not (:data (first causes2))))))
+
+(deftest compilation-errors-test
+  (is (match? #"Syntax error compiling at \(.*orchard/stacktrace_test\.clj:"
```
bbatsov, line 143: I prefer to have all assertions in `testing`, so that the output is a bit more descriptive, but it's not a big deal.

bbatsov:
Looks pretty good to me overall.

bbatsov:
Great work!

PR #323, opened 2025-03-13
[stacktrace] Stop sending compile-like flag

alexander-yakushev:
I found this confusing while I was redoing the the exception handling subsystem. This flag gives an impression that such exceptions (with "compile-like" being true) are handled differently by CIDER, e.g. they don't pop up cider-error buffer and only displayed inline with overlays. That's not true – this flag is not used anywhere in CIDER, cider-nrepl, or Calva. And I don't think it is a good idea to do that either – reflection errors are not compile-time errors and it's misleading to present them so.

To avoid the confusion, I propose to remove this unused flag. Changelog update does not feel necessary since `orchard.stacktrace` is still new.

### DIFF
--- a/src/orchard/stacktrace.clj
+++ b/src/orchard/stacktrace.clj
@@ -253,16 +253,6 @@
         (flag-duplicates)
         (flag-tooling))))
 
-(defn- compile-like-exception?
-  "'Compile-like' exceptions are those that happen at runtime (and therefore never
-  include a `:phase`) which however, represent code that cannot possibly work,
-  and thus are a compile-like exception (i.e. a linter could have caught them)."
-  [{:keys [phase type message]}]
-  (boolean
-   (and (nil? phase)
-        (= type 'java.lang.IllegalArgumentException)
-        (some->> message (re-find #"^No matching (field|method)")))))
-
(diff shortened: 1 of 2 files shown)


bbatsov:
I recall this was added as part of the changes to support Clojure 1.10 error data, but perhaps it became irrelevant at some point. Let me do a bit of digging in our archive to check what it was supposed to do.

bbatsov:
Took me a while to find it out, but it was introduced because of https://github.com/clojure-emacs/cider/pull/3423 

> are handled differently by CIDER, e.g. they don't pop up cider-error buffer and only displayed inline with overlays.

Well, the PR shows they were treated differently at some point. Now I have to check what happened with this configuration. 😅

alexander-yakushev:
I mean, this separation of compile-time and runtime exceptions is still there — it indeed works like that. But this "compile-like" is not looked at, only the presence of "phase".

bbatsov:
Ah, yeah. Now I see what you mean - I see that in CIDER's tests there are a couple of mentions of `compile-like`, but it's indeed not check for anywhere in the actual code. Seems this can be removed.

bbatsov (approved):

PR #325, opened 2025-03-19
[stacktrace] Add ex-str-formatted message to analyzed causes

alexander-yakushev:
To accommodate for changes I'm making in https://github.com/clojure-emacs/cider/pull/3789, we need the triaged compiler exception string to be available within the analyzed exception itself, not just the stderr output. The solution is clunky, and I would do it differently if we didn't already have this cruft of sending one message per cause, but it's cheaper from the compatibility perspective to add it here.

The second commit is a minor addition. The new pmap in orchard.misc is more efficient and doesn't yank the Futures pool.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#325](https://github.com/clojure-emacs/orchard/pull/325): Add `ex-str`-formatted message to analyzed causes.
+
 ## 0.31.0 (2025-03-14)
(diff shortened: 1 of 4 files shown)


alexander-yakushev:
Update: I also added the commit that removes printing customization from `orchard.stacktrace` and instead prints stuff with `orchard.print` that truncates long collections.

alexander-yakushev:
Merging to be able to review user-visible changes faster.

PR #331, opened 2025-04-04
[inspect] Add :table view-mode

alexander-yakushev:
Another big feature for the inspector that has been requested before and will make some people's lives better.

We already have the concept of a "view mode" that currently switches between :normal and :object (force-showing Java object fields instead of custom rendering).

A new view-mode called `:table` can also be cycled through using the same `v` keybinding. 

- Table view mode works for sequences of maps or tuples.
- Supports paging.
- All elements inside the table are clickable/navigable!
- There are some minor lackings, like path is not recorded if you navigate through the table, but it can be fixed later.

Here's how the end result looks:

image

image

image

---


bbatsov (approved):
I love this!

Great work! 🙇‍♂️

PR #333, opened 2025-04-07
[profile] Port thunknyc/profile to Orchard

alexander-yakushev:
As another milestone in my crusade against abandonware in CIDER, `orchard.profile` is intended to replace https://github.com/thunknyc/profile. This is a from-scratch implementation, so not much attribution is given. I dropped the parts of the API that cider-nrepl doesn't need; and users don't need them too because users should use a proper profiler, not this ersatz. Still, a few people in the survey reported using `cider-profile`, so let them be enjoy this new version.

As a way to display the results, I'm going with the new table view-mode just added to the inspector. Here's how the results will look like:

image

This will allow us to drop the custom `*cider-profile*` buffer, reuse the existing inspector machinery, give users the ability to click around, and even inspect the full list of timing samples (if anyone ever needs that).

---


bbatsov:
Impressive! I like how simple and short the new implementation is - only 188 lines!

bbatsov (approved):
Don't forget to mention it in the list of features in the README.

PR #335, opened 2025-04-11
Pretty view mode

r0man:
Hello,

this PR adds the pretty printer from https://github.com/eerohele/pp (as suggested by @alexander-yakushev) and some of its tests, and a configuration option to toggle the pretty printing with the "P" key binding in the CIDER inspector. 

The top level keys are rendered in the "inspector mode" as I understood it from a conversation on Slack.

The changes needed for the middleware can be found here: https://github.com/clojure-emacs/cider-nrepl/pull/932
The changes for CIDER are here: https://github.com/clojure-emacs/cider/pull/3813

The data structure shown on the following pictures was this:

```
{{:a 0
  :bb "000"
  :ccc []
  :d [{:a 0 :bb "000" :ccc []}
      {:a -1 :bb "111" :ccc [1]}
      {:a -2 :bb "222" :ccc [2 1]}
      {:a -3 :bb "333" :ccc [3 2 1]}
      {:a -4 :bb "444" :ccc [4 3 2 1]}]}
 {:a -1
  :bb "111"
  :ccc [1]
  :d [{:a 0 :bb "000" :ccc []}
      {:a -1 :bb "111" :ccc [1]}
      {:a -2 :bb "222" :ccc [2 1]}
      {:a -3 :bb "333" :ccc [3 2 1]}
      {:a -4 :bb "444" :ccc [4 3 2 1]}]}}
```

Without pretty printing enabled (as before):

map-normal-mode

With pretty printing enabled (in normal mode):

pprint-normal-mode

With pretty printing enabled (in object mode):

pprint-object-mode

Wdyt?

Open questions:
- Pretty printing applies only to collections as of now. Are there other objects except from collections that should be pretty printed differently? 
- I removed many of the pp tests and now kept only some of them in the `pp_test.clj` file. 
- I noticed the "Value" in :object mode did also got pretty printed with this change, but the indentation was off, so I adjusted this as well.
- Huge maps can look bad even in pretty mode. I tried it with one of our component system maps at work. I think the rendering could be improved if the content of a record (the `{...}` map after the record name) would be printed on a new line. Was the plan for the pretty printer to modify it for our needs? 

TODOs:
- I added 2 tests to `orchard.print-test`. They are passing in the REPL but fail when I run them with `lein test orchard.print-test` in Leiningen. For some reason they are printed differently in Leiningen. Any idea why that is? The atom `(atom 1)` for example is printed as `"#atom[1 0x4f9ea948]"` in the REPL, but like this `#object[clojure.lang.Atom 0x3a718b52 {:status :ready, :val 1}]` from Leiningen. I already checked that the various `*print-XXX*` vars are bound to the same value. It seems they are. What else could affect the printing?

---

- [ ] All tests are passing


src/orchard/inspect.clj
```diff
@@ -356,18 +372,68 @@
...
+  containing newlines."
+  [^String s]
+  (or (.contains s "\n") (> (count s) 20)))
+
+(defn- render-map-separator
+  "Render the map separator according to `rendered-key`. If
+  `rendered-key` is long or contains newlines the key and value will
+  be rendered on separate lines."
+  [{:keys [view-mode] :as inspector} rendered-key]
+  (if (and (= :pretty view-mode) (long-map-key? rendered-key))
```
alexander-yakushev, line 403: This and the few other predicates better be checked once and passed to these functions as an argument.

src/orchard/inspect.clj
```diff
@@ -356,18 +372,68 @@
...
+        (render-indent "=")
+        (render-ln))
+    (render inspector " = ")))
+
+(defn- render-map-value
+  "Render a map value. If `mark-values?` is true, attach the keys to the
+  values in the index."
+  [{:keys [view-mode] :as inspector} key val mark-values? rendered-key]
+  (if (= :pretty view-mode)
+    (let [indentation (if (long-map-key? rendered-key) 0 (+ 3 (count rendered-key)))]
```
alexander-yakushev, line 414: Ditto.

test/orchard/pp/pp_array_test.clj
```diff
@@ -0,0 +1,16 @@
+(ns orchard.pp.pp-array-test
```
alexander-yakushev, line 1: I don't think we need all pp tests in Orchard. Just 1-2 sanity checks for `orchard.print/pprint-str` is sufficient.

src/orchard/inspect.clj
```diff
@@ -356,18 +372,68 @@
         (render-ln))
     inspector))
 
+(defn- rendered-element-str
```
alexander-yakushev, line 375: I really don't like having to inspect what we have rendered after the fact and parsing it. It is preferable to orchestrate everything beforehand so that we don't need to do this.

alexander-yakushev (changes requested):

src/orchard/inspect.clj
r0man: I'm actually rendering the key before I render the separator and the map value (which currently need to know if the key spans multiple lines or not). What would be your suggestion instead? I kind of have to render the key to find out if it is a long key spanning multiple lines, no?

In another attempt, I looked back what we rendered from within `render-map-separator` and `render-map-value`, but I thought that is a bit more fragile.

alexander-yakushev: Yeah, rendering the key first is fine, but I suggest dragging around the information about being too long explicitly, not peek into the last value in `:rendered` to figure it out.

src/orchard/inspect.clj
r0man: I changed this now, and pass the result around.

test/orchard/pp/pp_array_test.clj
r0man: I removed most of the pp test files and combined some of the tests in `pp_test.clj` so we have some coverage for the `pp.clj` file as well. If you want me to remove those tests as well, let me know.

I also added 2 tests for `orchard.print/pprint-str`, but I have trouble getting them to run in Leiningen. For some reason the tests are passing in the REPL, but fail when I run them in Leiningen. atoms for example are printed differently. I mentioned this in the TODO section above. Any idea why that is?

src/orchard/inspect.clj
r0man: I changed this and pass the values I need around.

src/orchard/inspect.clj
```diff
@@ -44,13 +44,22 @@
...
 (defn- reset-render-state [inspector]
   (-> inspector
       (assoc :counter 0, :index [], :indentation 0, :rendered [])
       (dissoc :chunk :start-idx :last-page)))
 
+(defn- print-string
+  "Print or pretty print the string `value`, depending on the view mode
+  of the inspector."
+  [{:keys [indentation pretty-print]} value]
+  (if (= "true" pretty-print)
```
alexander-yakushev, line 59: It's better to use a normal boolean here, and convert the string at cider-nrepl side.

src/orchard/inspect.clj
```diff
@@ -220,6 +229,7 @@
   (when (some? max-nested-depth) (pre-ex (pos-int? max-nested-depth)))
   (when (some? display-analytics-hint) (pre-ex (= display-analytics-hint "true")))
   (when (some? analytics-size-cutoff) (pre-ex (pos-int? analytics-size-cutoff)))
+  (when (some? pretty-print) (pre-ex (#{"true" "false"} pretty-print)))
```
alexander-yakushev, line 232: Ditto re boolean.

src/orchard/inspect.clj
```diff
@@ -340,11 +354,15 @@
...
       (render-ln)))
 
-(defn render-labeled-value [inspector label value & [value-opts]]
-  (-> inspector
-      (render-indent (str label ": "))
-      (render-value value value-opts)
-      (render-ln)))
+(defn render-labeled-value [{:keys [pretty-print] :as inspector} label value & [value-opts]]
+  (let [formatted-label (str label ": ")
+        indentation (if (= "true" pretty-print) (count formatted-label) 0)]
```
alexander-yakushev, line 359: boolean

src/orchard/inspect.clj
```diff
@@ -356,18 +374,69 @@
...
+    (-> (render-ln inspector)
+        (render-indent "=")
+        (render-ln))
+    (render inspector " = ")))
+
+(defn- render-map-value
+  "Render a map value. If `mark-values?` is true, attach the keys to the
+  values in the index."
+  [{:keys [pretty-print] :as inspector} key val mark-values? rendered-key long-key?]
+  (if (= "true" pretty-print)
```
alexander-yakushev, line 415: boolean

src/orchard/inspect.clj
```diff
@@ -356,18 +374,69 @@
...
   [inspector mappable mark-values?]
   (reduce (fn [ins [key val]]
-            (-> ins
-                (render-indent)
-                (render-value key)
-                (render " = ")
-                (render-value val (when mark-values?
-                                    {:value-role :map-value, :value-key key}))
-                (render-ln)))
+            (let [rendered-key (rendered-element-str (last (:rendered (render-map-key ins key))))
```
alexander-yakushev, line 433: Still not quite what I meant:) You can print the key with `print-string` first. That gives you a string that you can check if it's long or has newlines and you don't have to dig it out of `:rendered` list and parse it. And once you have a printed string, you can pass it to `render-value` via `:display-value`.

alexander-yakushev:
Another thing: even though it is a config option now, I think it will still be beneficial to have a `toggle-pretty-print` function similar to `toggle-view-mode`, so that we can then also have a toggle op on cider-nrepl and thus on CIDER we can have a toggle command that doesn't need to know on Elisp side whether the pretty-printing is currently enabled.

.clj-kondo/config.edn
```diff
@@ -3,6 +3,18 @@
...
+                                                    non-map-coll-map-entry
+                                                    pp-vs-cpp-map-entry
+                                                    pp-vs-cpp-vec
+                                                    pp-vs-cpp-map-1
+                                                    pp-vs-cpp-map-2
+                                                    pp-vs-cpp-map-3
+                                                    pp-vs-cpp-inf-max-width
+                                                    print-readably
+                                                    print-arrays
+                                                    print-linear]}
```
alexander-yakushev, line 18: I suppose these are not needed anymore?

project.clj
```diff
@@ -5,7 +5,8 @@
 
 (def dev-test-common-profile
   {:dependencies '[[nubank/matcher-combinators "3.9.1"
-                    :exclusions [org.clojure/clojure]]]
+                    :exclusions [org.clojure/clojure]]
+                   [org.clojure/test.check "1.1.1"]]
```
alexander-yakushev, line 9: Ditto.

alexander-yakushev:
Whoops, sorry for confusing you! I forgot that `toggle-view-mode` is implemented completely on cider-nrepl side. Right, it is enough to have `:pretty-print` as a configuration option, and then implement `toggle-pretty-print` in cider-nrepl since it has complete access to inspector state anyway. Sorry again.

alexander-yakushev:
I think it is perfect as it is. Let's merge!

alexander-yakushev (approved):

PR #337, opened 2025-04-22
[print] Make orchard.print more consistent with CIDER printing

alexander-yakushev:
I want to make the custom printing we do in inspector, debugger, and the customly inject print-methods in cider-nrepl more consistent with each other. This PR adds print handling for futures, promises, namespaces, multimethods (things that cider-nrepl already performs prettyfication of), downcases the custom printed representation names (again, like cider-nrepl does), and other minor things.

---


bbatsov (approved):
The proposed changes look good to me.

r0man:
Nice! Thanks for fine tuning this! :heart:

PR #338, opened 2025-04-22
[pp] Reuse orchard.print in orchard.pp

alexander-yakushev:
This achieves better displayed consistency between orchard.print and orchard.pp. It also cuts down some code from orchard.pp that's already implemented in orchard.print.

---

### DIFF
--- a/src/orchard/pp.clj
+++ b/src/orchard/pp.clj
@@ -1,5 +1,8 @@
 (ns orchard.pp
-  "A pretty-printer for Clojure data structures.
+  "A pretty-printer for Clojure data structures. This namespace is borrowed from
+  Eero Helenius' pp project and modified according to Orchard needs. Linear
+  printing parts were replaced with substitutes from `orchard.print` for reuse
+  and consistency.
 
   Based on the algorithm described in \"Pretty-Printing, Converting List
   to Linear Structure\" by Ira Goldstein (Artificial Intelligence, Memo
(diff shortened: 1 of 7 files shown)


src/orchard/pp.clj
```diff
@@ -1,13 +1,14 @@
 (ns orchard.pp
   "A pretty-printer for Clojure data structures.
-
-  Based on the algorithm described in \"Pretty-Printing, Converting List
-  to Linear Structure\" by Ira Goldstein (Artificial Intelligence, Memo
-  No. 279 in Massachusetts Institute of Technology A.I. Laboratory,
-  February 1973)."
+  This namespace is borrowed from Eero Helenius' pp project and modified
```
bbatsov, line 3: I think it's still a good idea to leave the reference to the underlying algorithm, as I doubt that everyone will go and check the upstream repo.

bbatsov: (and perhaps high-level differences with the original implementation)

bbatsov (approved):
The changes look good to me overall.

src/orchard/pp.clj
```diff
@@ -150,149 +137,17 @@
...
-
-    :level (long)
-      The current nesting level."
-  ([form]
-   (print-linear form nil))
-  (^String [form opts]
-   (with-str-writer (fn [writer] (print-linear writer form opts))))
-  ([writer form opts]
-   (-print form writer opts)))
+(defn- print-str-linear ^String [x opts]
```
bbatsov, line 140: I'd probably keep some of the original docstring here.

PR #340, opened 2025-04-28
[ci] Run Eastwood together with other linters

alexander-yakushev:
We currently only test with JDK8 and 24, and with 8 being soft-deprecated anyway, it makes sense to just treat Eastwood as all other linters.


bbatsov (approved):
Agreed!

PR #342, opened 2025-05-26
[inspect] Hexdump view mode

alexander-yakushev:
Here is how it looks:

image

- Inspired by `hexdump` tool.
- Supports paging (renders 16*page-size of bytes).
- Currently only works for byte arrays. I may add support for ByteBuffers in the future, based on the feedback.

---

### DIFF
--- a/src/orchard/inspect.clj
+++ b/src/orchard/inspect.clj
@@ -35,7 +35,7 @@
                             (list 'get key)))
     (conj path '<unknown>)))
 
-(def ^:private supported-view-modes #{:normal :object :table})
+(def ^:private supported-view-modes #{:normal :object :table :hex})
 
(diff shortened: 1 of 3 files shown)


src/orchard/inspect.clj
```diff
@@ -655,6 +656,45 @@
       (unindent ins))
     inspector))
 
+;; Hex view mode
+
+(defn- byte->ascii [b]
+  (let [c (bit-and b 0xFF)]
+    (if (and (>= c 32) (<= c 126))
+      (char c)
+      ;; Use MIDOT ABOVE for non-printed chars as it is distinct from 0x2E.
```
bbatsov, line 665: MIDOT ABOVE?

PR #344, opened 2025-05-28
[inspect] Custom messages for inspecting nil

alexander-yakushev:
This is just a small bit of fun and diversion.

image

---

- [x] All tests are passing


bbatsov (approved):
I love it! That's the spirit of CIDER! :D

PR #347, opened 2025-06-17
[inspect] Performance refactoring

alexander-yakushev:
More inspector refactoring with no significant user-facing component but unifies some things and improves performance (useful when e.g. flipping rapidly through a paginated collection).


bbatsov (approved):
Looks solid to me!

PR #382, opened 2026-04-07
Rewrite orchard.xref

alexander-yakushev:
Big redoing of `orchard.xref` that includes:

- Improved performance.
- Getting rid of `eval` (yikes).
- Cleanup and compaction.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- [#382](https://github.com/clojure-emacs/orchard/pull/382): Rewrite and optimize `orchard.xref`.
+
 ## 0.40.0 (2026-03-25)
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks for looking into this! How did you even think of revisiting it? 

I was wondering if the past if this is useful in its current state or no (and even considered replacing it with some xref that just uses regular expressions like swank-clojure did), but I guess it's better to have it than not to have it, and we're going to have it - it's better to optimize it. :-)

alexander-yakushev:
Just was cruising through the repo and found a namespace I haven't touched before:).

I'm not sure if anybody ever uses it, but now at least it's in a state that is not embarrassing to keep:).

bbatsov (approved):

PR #396, opened 2026-05-20
Indent: unwrap (quote ...) around :arglists

bbatsov:
Fixes the orchard side of clojure-emacs/cider#3923.

`infer-style-indent` destructures `:arglists` as a list of arglist vectors. ClojureScript's analyzer doesn't always hand it over that way: when `:arglists` comes from a macro-written macro (one whose body emits `'([x])` rather than a bare `([x])` for the arglists meta), the analyzer preserves the raw form including the `quote`. So the value arrives as `(quote ([ctx & body]))` and the destructuring binds the first element of that, which is the symbol `quote`, not a list. `find-idx` then calls `.indexOf` on a symbol and we get a ClassCastException.

The trace from the original issue:

```
java.lang.ClassCastException: class clojure.lang.Symbol cannot be cast to class java.util.List
        at orchard.indent$find_idx.invokeStatic(indent.clj:107)
        at orchard.indent$compute_style_indent.invokeStatic(indent.clj:120)
        at orchard.indent$infer_style_indent.invokeStatic(indent.clj:164)
        at cider.nrepl.middleware.track_state$ns_state$post_process__64337$fn__64338.invoke(track_state.clj:217)
```

This propagates up through cider-nrepl's track-state middleware and tanks `cider/get-state` for the whole connection, which is the user-visible symptom: affected buffers can't link to their REPL session.

The fix peels the `quote` wrapper off in a small `normalize-arglists` helper applied once at the entry point. No change for the common case (bare arglists, no quote wrapper). Tests cover both the unwrapping path and the pass-through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 - [#386](https://github.com/clojure-emacs/orchard/pull/386): Inspector: rename default view mode for unknown objects to `:object`.
+- [#396](https://github.com/clojure-emacs/orchard/pull/396): Indent: don't crash on `(quote (...))`-wrapped `:arglists` (clojure-emacs/cider#3923).
 - [#397](https://github.com/clojure-emacs/orchard/pull/397): Inspector: add support for diffing sets.
 
 ## 0.41.0 (2026-04-13)
--- a/src/orchard/indent.clj
+++ b/src/orchard/indent.clj
@@ -148,6 +148,20 @@
              'fntail
              'fn-tail]))))
 
+(defn- normalize-arglists
+  "Unwrap a `(quote ...)` wrapper around `arglists`, if present.
+
+  ClojureScript's analyzer preserves `:arglists` as the raw form when it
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 - [#386](https://github.com/clojure-emacs/orchard/pull/386): Inspector: rename default view mode for unknown objects to `:object`.
+- Indent: stop crashing on `(quote (...))`-wrapped `:arglists`. ClojureScript's analyzer preserves the raw form (including the `quote`) when `:arglists` comes from a macro-written macro, so `infer-style-indent` would hit a `ClassCastException` trying to treat the `quote` symbol as a list. See clojure-emacs/cider#3923.
```
alexander-yakushev: The CHANGELOG entry looks more verbose than what we usually have. All those details can live perfectly fine behind the link.

alexander-yakushev (approved):
Approved with a nitpick.

PR #401, opened 2026-06-24
Add protocol-impls and multimethod-dispatch-values

bbatsov:
The foundational piece for a "who-implements" feature in CIDER (clojure-emacs/cider#1969, whose original design sketch called for exactly this middleware).

`orchard.xref/protocol-impls` returns the types implementing a protocol. It covers both `extend`/`extend-type`/`extend-protocol` targets (via `extenders`) and inline `defrecord`/`deftype` implementers - the latter found by scanning the loaded-class cache for the protocol's generated interface (the same `DynamicClassLoader` cache `fn-deps` already uses). Each entry carries a source location when one is resolvable (records/types expose it through their `->Name` factory var; Java targets have none).

`orchard.xref/multimethod-dispatch-values` returns a multimethod's dispatch values.

Only loaded code is visible, consistent with the rest of `orchard.xref`. Per-`defmethod` source locations aren't included - those methods are anonymous fns with no runtime metadata, so they'd need bytecode line-table reading; that's left for a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
+- [#401](https://github.com/clojure-emacs/orchard/pull/401): Add `orchard.xref/protocol-impls` (the types implementing a protocol, including inline `defrecord`/`deftype` implementers, with source locations) and `orchard.xref/multimethod-dispatch-values`.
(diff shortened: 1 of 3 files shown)


harold:
Super :cool: !

PR #404, opened 2026-07-02
[xref] Small improvements to protocol discovery

alexander-yakushev:
The code NPEs on certain clojure-core protocols, ~also remove a try/catch that should never fire.~

Holy crap, it does fire. Slightly prettify anyway.


bbatsov:
Should we also add a couple of tests for the fix?

alexander-yakushev:
Added `nil` extender to TestShape, so now it tests for nils too.

bbatsov (approved):

PR #406, opened 2026-07-03
[inspector] Remove datafy section

alexander-yakushev:
Less is more.

---


alexander-yakushev:
I'm going to merge this and proceed with follow-up work. If we are to bring it back at some point, it will have a different form.

bbatsov:
No worries. My availability will be somewhat limited the next few days, so no reason to wait on me.

PR #417, opened 2026-07-15
[pp] Honor custom print-method for Clojure collection types

bbatsov:
Follow-up to #415. `orchard.pp` has its own structural dispatch (the `PrettyPrintable` protocol), so the pretty-print view still traversed records, datasets and collection deftypes that define their own `print-method`.

Now the collection extensions consult `orchard.print/custom-print-method?` (made public here) and print the custom representation instead. The interesting part is layout: a real `tech.ml.dataset` prints a multi-line table, so the custom output is written line by line to keep the writer's column accounting and indentation intact - the `Object` fallback gets the same treatment. `*print-level*` and `*print-meta*` behave as before, and if a custom `print-method` throws, we fall back to the regular structural pretty-printing.


bbatsov:
Closing this for now. Following the discussion around #415 we're rethinking whether honoring print-method is the right default at all, and pp staying structural matches what clojure.pprint does anyway. I'll extract the unrelated multi-line fix for the Object fallback into a separate PR.

PR #420, opened 2026-08-03
[ci] Tune test matrices

alexander-yakushev:
Since Clojure 1.13 is now only supported with JDK17+, I've rotated test matrices a bit. Now we have:
- stable Clojure (1.12) tested with all supported JDKs
- alpha Clojure (1.13) tested with jdk17+
- previous Clojure versions sanity-tested with oldest jdk (8) and latest (26).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -150,7 +150,7 @@ jobs:
                 name: Running tests (make << parameters.command >>)
                 command: make << parameters.command >>
 
-  test_windows:
+  test-windows:
     description: |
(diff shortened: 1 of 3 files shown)


bbatsov:
I'm fine with the proposal, but perhaps we should also add your reasoning as comments to the matrix itself, so it's clearer why we've organized it in this manner. (otherwise some people might wonder about the difference between Clojure 1.11 and 1.12 for instance)

alexander-yakushev:
Updated. Something like this?

PR #428, opened 2026-09-17
[java] Release the file manager and remove copied sources after each parse

sundbp:
`orchard.java.parser-next/parse-java` creates a temporary directory, copies the Java source into it, and opens a `StandardJavaFileManager`. The old `finally` tried to delete only the non-empty directory, so the copied source survived every call. The file manager used to enumerate the copied source was also never closed.

Observed in practice: CIDER parses `mx.cider.orchard.LruMap` while loading its namespace, so every REPL session that loads CIDER leaves one `tmp*/LruMap.java` behind.

This patch keeps the complete resource lifetime inside `parse-java`. The function now accepts a callback that consumes the `DocletEnvironment` while its backing resources are valid. It creates one `StandardJavaFileManager`, passes that manager to `getTask`, invokes the callback, closes the manager with `with-open`, and then removes the copied source and temporary directory in `finally`. `source-info` supplies its existing parse-tree traversal as that callback.

The existing `parser-next-test` namespace now includes one focused regression that observes the copied source created by the parser and verifies that both the file and its directory are gone after `source-info` returns. The existing invalid-source test was updated for the private callback argument.

Locally verified on Temurin JDK 26 / Clojure 1.12.5: successful source parsing returns the expected class information and removes both temporary artifacts; invalid source parsing preserves the existing diagnostic and also removes both artifacts. `cljfmt` and `clj-kondo` pass.


alexander-yakushev:
This looks like an overly complicated fix to a minor problem, and +300 is certainly too much. If you condense it to something trivial, I'll consider the fix.

sundbp-crucible:
The change for the fix is ~50 lines, the added tests are the main addition. I'll try to shrink it. My REPL sessions were generating hundreds and thousands of these tmp files, which is why I had a go at fixing the cleanup.

sundbp-crucible:
ok, managed to shrink it down and change existing tests instead. what do you think of this version?

